PR #87, opened 2017-07-24
Refactorization - Review

ethman:


nussl/separation/duet.py
```diff
@@ -113,102 +164,78 @@ def run(self):
...
 
+    def _compute_atn_delay(self):
         # Calculate the symmetric attenuation (alpha) and delay (delta) for each
         # time-freq. point
-        R21 = (stft_ch1 + nussl.constants.EPSILON) / (stft_ch0 + nussl.constants.EPSILON)
-        relative_attenuation = np.abs(R21)  # relative attenuation between the two channels
-        symmetric_attenuation = relative_attenuation - 1 / relative_attenuation
-        relative_delay = -np.imag(np.log(R21)) / (2 * np.pi * frequency_matrix)  # relative delay
+        inter_channel_ratio = (self.stft_ch1 + nussl.constants.EPSILON) / (self.stft_ch0 + nussl.constants.EPSILON)
+        atn = np.abs(inter_channel_ratio)  # relative attenuation between the two channels
```
ethman: `attenuation`

nussl/separation/duet.py
```diff
@@ -113,102 +164,78 @@ def run(self):
...
+        symmetric_atn = atn - 1 / atn  # symmetric attenuation
+        delay = -np.imag(np.log(inter_channel_ratio)) / (2 * np.pi * self.frequency_matrix)  # relative delay
```
ethman: `relative_delay`

nussl/separation/duet.py
```diff
@@ -113,102 +164,78 @@ def run(self):
...
+        return symmetric_atn, delay
+
+    def make_histogram(self, p, q):
```
ethman: Does this function need to be public? Would a user use this function by itself, or does it only work in the context of the rest of the code? If it's the latter, then make it private.

nussl/separation/duet.py
```diff
@@ -113,102 +164,78 @@ def run(self):
...
-        tfw = (np.abs(stft_ch0) * np.abs(stft_ch1)) ** p * (np.abs(frequency_matrix)) ** q  # time-freq weights
+        time_frequency_weights = (np.abs(self.stft_ch0) * np.abs(self.stft_ch1)) ** p \
+                                 * (np.abs(self.frequency_matrix)) ** q
 
         # only consider time-freq. points yielding estimates in bounds
-        a_premask = np.logical_and(self.attenuation_min < symmetric_attenuation, symmetric_attenuation < self.attenuation_max)
-        d_premask = np.logical_and(self.delay_min < relative_delay, relative_delay < self.delay_max)
+        a_premask = np.logical_and(self.attenuation_min < self.symmetric_atn, self.symmetric_atn < self.attenuation_max)
+        d_premask = np.logical_and(self.delay_min < self.delay, self.delay < self.delay_max)
         ad_premask = np.logical_and(a_premask, d_premask)
```
ethman: I'm not a big fan of a lot of these variable names here. `ad_nzind`, *`_vec`, `tfw_vec`, these are all unclear to me.

nussl/separation/duet.py
```diff
@@ -113,102 +164,78 @@ def run(self):
...
-        agrid = H[1]
-        dgrid = H[2]
-
         # Save these for later
-        self.attenuation_grid = agrid
-        self.delay_grid = dgrid
-        self.attenuation_delay_histogram = hist
-        self.non_normalized_hist = H[0]
-        self.smoothed_hist = self._smooth_matrix(self.non_normalized_hist, np.array([3]))
+        self.attenuation_grid = atn_bins
```
ethman: These names make sense, but why not save them as attributes when you get from from `np.histogram2d`?

nussl/separation/duet.py
```diff
@@ -113,102 +164,78 @@ def run(self):
...
+        self.delay_grid = delay_bins
 
-        # smooth the histogram - local average 3-by-3 neighboring bins
-        hist = self._smooth_matrix(hist, np.array([3]))
+        # TODO: do these operations happen twice?
```
ethman: Is this `TODO` resolved? If so, why is it still in the code?

nussl/separation/duet.py
```diff
@@ -113,102 +164,78 @@ def run(self):
...
+        # resolved
+        self.attenuation_delay_histogram = hist / hist.max()
```
ethman: This should be `attenuation_delay_histogram_normalized` and the other should be just `attenuation_delay_histogram`. Make sure the variable names are clear.

nussl/separation/duet.py
```diff
@@ -113,102 +164,78 @@ def run(self):
...
+        self.non_normalized_hist = hist
+
+        # smooth the normalized histogram - local average 3-by-3 neighboring bins
+        hist = self._smooth_matrix(self.attenuation_delay_histogram, np.array([3]))
```
ethman: Did we talk about this? Why is there both `hist` and `self.attenuation_delay_histogram`? **Don't rewrite code!!!** If you've already calculated the thing you need, use it, don't recalculate it!

nussl/separation/duet.py
```diff
@@ -113,102 +164,78 @@ def run(self):
...
-
-            xhat[i, :] = nussl.utils.add_mismatched_arrays(xhat[i,], xi)[:self.audio_signal.signal_length]
-            # add back to the separated signal a portion of the mixture to eliminate
-            # most of the masking artifacts
-            # xhat=xhat+0.05*x[0,:]
-
-        self.separated_sources = xhat
-
-        return xhat, ad_est
+        return hist, atn_bins, delay_bins
```
ethman: Do you need to return these things if they're already saved as attributes? Do you use the attributes or do you use the return values? Pick one.

nussl/separation/duet.py
```diff
@@ -113,102 +164,78 @@ def run(self):
...
 
     @staticmethod
     def find_peaks2(data, min_thr=0.5, min_dist=None, max_peaks=1):
```
ethman: Are you still using this in the code? If not, delete it.

nussl/separation/duet.py
```diff
@@ -254,6 +282,37 @@ def find_peaks2(data, min_thr=0.5, min_dist=None, max_peaks=1):
...
+       Parameters:
+            atn_bins (np.array): Attenuation bins
+            delay_bins (np.array): Delay bins
+
+        Returns:
+            delay_peak(np.array): The delay peaks determined from the histogram
+            atn_delay_est (np.array): The estimated symmetric attenuation and delay values
+            atn_peak (np.array): Attenuation converted from symmetric attenuation
+        """
+        # TODO:
```
ethman: What's this `TODO` for?

nussl/separation/duet.py
```diff
@@ -254,6 +282,37 @@ def find_peaks2(data, min_thr=0.5, min_dist=None, max_peaks=1):
 
         return peakIndex
 
+    def convert_peaks(self, atn_bins, delay_bins):
```
ethman: Same thing as above. Should this be a private function?

nussl/separation/duet.py
```diff
@@ -254,6 +282,37 @@ def find_peaks2(data, min_thr=0.5, min_dist=None, max_peaks=1):
...
+        delay_peak = delay_bins[delay_indices]
+
+        atn_delay_est = np.column_stack((symmetric_atn_peak, delay_peak))
+
+        # convert symmetric_atn to atn_peak
+        atn_peak = (symmetric_atn_peak + np.sqrt(symmetric_atn_peak ** 2 + 4)) / 2
+        return delay_peak, atn_delay_est, atn_peak
+
     @staticmethod
     def find_peaks(data, min_thr=0.5, min_dist=None, max_num=1):
```
ethman: Is this still used? If not, delete it.

nussl/separation/duet.py
```diff
@@ -254,6 +282,37 @@ def find_peaks2(data, min_thr=0.5, min_dist=None, max_peaks=1):
...
+        if any(a > len(atn_bins) for a in atn_indices):
+            raise ValueError('Attenuation index greater than length of bins')
+
+        symmetric_atn_peak = atn_bins[atn_indices]
+        delay_peak = delay_bins[delay_indices]
+
+        atn_delay_est = np.column_stack((symmetric_atn_peak, delay_peak))
+
+        # convert symmetric_atn to atn_peak
+        atn_peak = (symmetric_atn_peak + np.sqrt(symmetric_atn_peak ** 2 + 4)) / 2
```
ethman: Where do these values (`**2 + 4)) / 2`) come from? Cite them

nussl/separation/duet.py
```diff
@@ -297,6 +356,63 @@ def find_peaks(data, min_thr=0.5, min_dist=None, max_num=1):
 
         return peak_indices
 
+    def compute_masks(self, atn_peak, delay_peak):
```
ethman: Should this be private?

nussl/separation/duet.py
```diff
@@ -297,6 +356,63 @@ def find_peaks(data, min_thr=0.5, min_dist=None, max_num=1):
...
+        for i in range(0, self.num_sources):
+            phase = np.exp(-1j * self.frequency_matrix * delay_peak[i])
+            score = np.abs(atn_peak[i] * phase * self.stft_ch0 - self.stft_ch1) ** 2 / (1 + atn_peak[i] ** 2)
+            mask = (score < best_so_far)
+            best_ind[mask] = i
+            best_so_far[mask] = score[mask]
+
+        return best_ind, mask
+
+    def convert_time_domain(self, best_ind, mask, atn_peak, delay_peak):
```
ethman: Rename this so it's more descriptive. Convert to or from the time domain? Convert what? Also: should this be private?

nussl/separation/duet.py
```diff
@@ -297,6 +356,63 @@ def find_peaks(data, min_thr=0.5, min_dist=None, max_num=1):
...
+            best_ind (np.array): The indices for where the masks are applied
+            mask (np.array): binary mask used to separate the sources
+            atn_peak (np.array): Attenuation peaks determined from histogram
+            delay_peak (np.array): Delay peaks determined from histogram
+
+        Returns:
+           source_estimates (np.array): The arrays of the separated source signals
+
+        """
+        source_estimates = np.zeros((self.num_sources, self.audio_signal.signal_length))
```
ethman: Don't pre-allocate the numpy array. Start with a list and convert it to a numpy array when you're done.

nussl/separation/duet.py
```diff
@@ -297,6 +356,63 @@ def find_peaks(data, min_thr=0.5, min_dist=None, max_num=1):
...
+        for i in range(0, self.num_sources):
+            mask = (best_ind == i)
+            Xm = np.vstack([np.zeros((1, self.num_time_bins)),
```
ethman: `Xm` is a bad variable name. Also, get rid of this vstack (looping through a list might help here).

nussl/separation/duet.py
```diff
@@ -297,6 +356,63 @@ def find_peaks(data, min_thr=0.5, min_dist=None, max_num=1):
...
+            delay_peak (np.array): Delay peaks determined from histogram
+        Returns:
+           best_ind(np.array): The indices for where the masks are applied
+           mask (np.array): A binary mask used to separate the sources
+
+        """
+        # compute masks for separation
+        best_so_far = np.inf * np.ones((self.num_frequency_bins - 1, self.num_time_bins))
+        best_ind = np.zeros((self.num_frequency_bins - 1, self.num_time_bins), dtype=int)
+        mask = None
```
ethman: Right now this mask looks like this (in 1D): `[ 0, 0, 3, 2, 0, 2, 2, 1, 1, 3, 3 ]` where each number is a mask. This is not the most organized way to do this. I'm going to have to look at the new `MaskBase` methods and use those instead here. This method should return a list of `MaskBase`-derived objects instead of this Frankenstein mask array.

nussl/separation/duet.py
```diff
@@ -297,6 +356,63 @@ def find_peaks(data, min_thr=0.5, min_dist=None, max_num=1):
...
+        """
+        source_estimates = np.zeros((self.num_sources, self.audio_signal.signal_length))
+        for i in range(0, self.num_sources):
+            mask = (best_ind == i)
+            Xm = np.vstack([np.zeros((1, self.num_time_bins)),
+                            (self.stft_ch0 + atn_peak[i] * np.exp(1j * self.frequency_matrix * delay_peak[i]) * self.stft_ch1) /
+                            (1 + atn_peak[i] ** 2) * mask])
+
+            # xi = spectral_utils.f_istft(Xm, L, winType, hop, fs)
+            xi = nussl.spectral_utils.e_istft(Xm, self.stft_params.window_length,
```
ethman: Don't do this. Make a list of new `AudioSignal` copies (there are helper methods to do this) with the masked stft data and just call `AudioSignal.istft()`.

nussl/separation/duet.py
```diff
@@ -334,7 +450,7 @@ def _smooth_matrix(matrix, kernel):
         copy_row = int(np.floor(krow / 2))  # number of rows to copy on top and bottom
         copy_col = int(np.floor(kcol / 2))  # number of columns to copy on either side
 
-        # TODO: This is very ugly. Make this readable
+        # TODO: This is very ugly. Make this readable, write issue on Github
```
ethman: Is there an issue for this?

nussl/separation/duet.py
```diff
@@ -388,7 +504,10 @@ def plot(self, output_name, three_d_plot=False, normalize=True):
         AA = np.tile(self.attenuation_grid[1::], (self.num_delay_bins, 1)).T
         DD = np.tile(self.delay_grid[1::].T, (self.num_attenuation_bins, 1))
 
-        histogram_data = self.attenuation_delay_histogram if normalize else self.smoothed_hist
+        # save smoothed_hist (not normalized) as an option for plotting
+        smoothed_hist = self._smooth_matrix(self.non_normalized_hist, np.array([3]))
```
ethman: Where does this `np.array([3])` come from? You should change `self._smooth_matrix` so that it doesn't take a numpy array of a single element. (Unless it needs to, in which case you need to justify that and explain what's going on here.) Also, why 3?

nussl/separation/duet.py
```diff
@@ -388,7 +504,10 @@ def plot(self, output_name, three_d_plot=False, normalize=True):
         AA = np.tile(self.attenuation_grid[1::], (self.num_delay_bins, 1)).T
```
ethman: What're `AA` and `DD`? And also what's this syntax: `[1::]`?

nussl/separation/duet.py
```diff
@@ -71,15 +76,24 @@ def __init__(self, input_audio_signal, num_sources,
 
         self.attenuation_delay_histogram = None
         self.non_normalized_hist = None
-        self.smoothed_hist = None
         self.peak_indices = None
+        self.p = p
```
ethman: If you can't find a better name for these in the paper, then add a comment so that it's clear what they do. Or make up a descriptive name.

nussl/audio_signal.py
```diff
@@ -207,14 +222,45 @@ def audio_data(self, value):
...
 
-        if self._audio_data.ndim < 2:
-            self._audio_data = np.expand_dims(self._audio_data, axis=constants.CHAN_INDEX)
+        self._audio_data = value
 
         self.set_active_region_to_default()
 
     @property
+    def stft_data(self):
+        """ (:obj:`np.ndarray`): Real-valued, uncompressed, time-domain representation of the audio.
```
ethman: This comment is copied and pasted from `audio_data`. Make sure you update this,

nussl/audio_signal.py
```diff
@@ -1157,7 +1220,7 @@ def get_magnitude_spectrogram_channel(self, n):
         # np.array helps with duck typing
         return utils._get_axis(np.array(self.magnitude_spectrogram_data), constants.STFT_CHAN_INDEX, n)
 
-    def to_mono(self, overwrite=False):
+    def to_mono(self, overwrite=False, remove_channels=True):
```
ethman: What does this new parameter do again? I forget. We'll talk about if this is still needed, but you should at the very least document what you added. Make this a habit: write code --> write documentation!

nussl/audio_signal.py
```diff
@@ -1170,11 +1233,32 @@ def to_mono(self, overwrite=False):
...
+            mono = np.mean(self.audio_data, axis=constants.CHAN_INDEX, keepdims=True)
         if overwrite:
             self.audio_data = mono
         return mono
 
+    def stft_to_mono(self, overwrite=False):
+        """ Converts :attr:`stft_data` to mono by averaging every sample.
+
+        Warning:
+            If overwrite=True (default) this will overwrite any data in :attr:`audio_data`!
```
ethman: typo: `audio_data` --> `stft_data`

nussl/audio_signal.py
```diff
@@ -1170,11 +1233,32 @@ def to_mono(self, overwrite=False):
...
+        if remove_channels:
+            mono = np.mean(self.audio_data, axis=constants.CHAN_INDEX)
+        else:
+            mono = np.mean(self.audio_data, axis=constants.CHAN_INDEX, keepdims=True)
         if overwrite:
             self.audio_data = mono
         return mono
 
+    def stft_to_mono(self, overwrite=False):
+        """ Converts :attr:`stft_data` to mono by averaging every sample.
```
ethman: Hmm... maybe specify that it will be one channel. I'm used to seeing the term 'mono' used in the time domain, but not in the time-frequency domain. Also, what will the shape of `stft_data` be after you run this function (Hint: `[NUM_FREQ, NUM_HOPS, 1]`)?

nussl/separation/duet.py
```diff
@@ -0,0 +1,440 @@
...
+import nussl.constants
+import nussl.spectral_utils
+import nussl.utils
+import mask_separation_base
+import masks
+
+
+class Duet(mask_separation_base.MaskSeparationBase):
+
+    """Implements the Degenerate Unmixing Estimation Technique (DUET) algorithm.
```
ethman: Overall this documentation is a lot nicer. Good job beefing it up. One thing I think that could help the docs a lot is if you gave a brief overview about how the algorithm works (feel free to use math and/or code examples from this file). Just to add context to what 'attenuation' and 'delay' are. Add a paragraph or so, not too much so that a user gets the gist and can make intelligent decisions about how to set the parameters.

nussl/separation/duet.py
```diff
@@ -0,0 +1,440 @@
...
+        attenuation_min (int): Lower bound on attenuation value, used as minimum distance in utils.find_peak_indices.
+        attenuation_max (int): Upper bound on attenuation, used for creating a histogram without outliers.
+        num_attenuation_bins (int): Number of bins for attenuation. 
+        delay_min (int): Lower bound on delay, used as minimum distance in utils.find_peak_indices.
+        delay_max (int): Upper bound on delay, used for creating a histogram without outliers.
+        num_delay_bins (int): Number of bins for delay. 
+        peak_threshold (float): Value in [0, 1] for peak picking. 
+        attenuation_min_distance (int): Minimum distance between peaks wrt attenuation. 
+        delay_min_distance (int): Minimum distance between peaks wrt delay.
+        p (int): Weight the histogram with the symmetric attenuation estimator
```
ethman: I remember the paper had some recommendations for values for these two variables. Could you reproduce those recommendations here? (Make sure to cite the paper if you decide to directly copy/paste, else paraphrasing is fine.)

ethman: "These values" == `p` and `q`.

nussl/separation/duet.py
```diff
@@ -0,0 +1,440 @@
...
+        q (int): Weight the histogram with the delay estimator
+
+
+    Attributes:
+        stft_ch0 (np.array): A Numpy matrix containing the stft data of channel 0.
+        stft_ch1 (np.array): A Numpy matrix containing the stft data of channel 0.
```
ethman: Typo: "Channel 1"

nussl/separation/duet.py
```diff
@@ -0,0 +1,440 @@
...
+    Examples:
+        :ref:`The DUET Demo Example <duet_demo>`
+
+    """
+
+    def __init__(self, input_audio_signal, num_sources,
+                 attenuation_min=-3, attenuation_max=3, num_attenuation_bins=50,
+                 delay_min=-3, delay_max=3, num_delay_bins=50,
+                 peak_threshold=0.2, attenuation_min_distance=5, delay_min_distance=5, p=1, q=0,
+                 mask_type=mask_separation_base.MaskSeparationBase.BINARY_MASK, mask_threshold=0.5):
```
ethman: Don't expose `mask_type` and `mask_threshold` because DUET only produces binary masks. So get rid of these as parameters to `__init__()` and when you call the constructor of the super class (in line 74 below), just set `mask_type=BINARY_MASK` (or whatever the syntax) and no need to set anything for `mask_threshold` just leave it out.

nussl/separation/duet.py
```diff
@@ -0,0 +1,440 @@
...
+        Example:
+        .. code-block:: python
+            :linenos:
+
+            #Import input audio signal
+            input_file_name = '../Input/dev1_female3_inst_mix.wav'
+            signal = AudioSignal(path_to_input_file=input_file_name)
+
+            # Set up and run Duet
+            duet = Duet(signal, a_min=-3, a_max=3, a_num=50, d_min=-3, d_max=3, d_num=50, threshold=0.2,
```
ethman: Make sure this example is up to date.

nussl/separation/duet.py
```diff
@@ -0,0 +1,440 @@
...
+        n_time_bins = len(self.audio_signal.time_bins_vector)
+        wmat = np.array(np.tile(np.mat(self.audio_signal.freq_vector).T, (1, n_time_bins))) * (2 * np.pi / sample_rate)
+        wmat += nussl.constants.EPSILON
+        return stft_ch0, stft_ch1, wmat
+
+    @staticmethod
+    def _compute_atn_delay(stft_ch0, stft_ch1, frequency_matrix):
+        # Calculate the symmetric attenuation (alpha) and delay (delta) for each
+        # time-freq. point
+        inter_channel_ratio = (stft_ch1 + nussl.constants.EPSILON) / (stft_ch0 + nussl.constants.EPSILON)
```
ethman: Hmmm... Looks like this happens twice (the first is up at line 192-3). I know I told you to do the first one. Can you test what happens if you get rid of the epsilons at 192-3. Just curious. We'll have a chat about it soon.

nussl/separation/duet.py
```diff
@@ -0,0 +1,440 @@
...
+
+        if any(a > len(self.attenuation_bins) for a in atn_indices):
+            raise ValueError('Attenuation index greater than length of bins')
+
+        symmetric_atn_peak = self.attenuation_bins[atn_indices]
+        delay_peak = self.delay_bins[delay_indices]
+
+        atn_delay_est = np.column_stack((symmetric_atn_peak, delay_peak))
+
+        # convert symmetric_atn to atn_peak using formula from Rickard
```
ethman: Can you add an equation/page number from the paper?

nussl/separation/duet.py
```diff
@@ -0,0 +1,440 @@
...
+                self.masks[i])  # Ask Ethan if we can have a apply_mask that doesn't return a new mask!
+            new_sig.stft_params = self.stft_params
+            source_estimate = new_sig.istft(overwrite=True, truncate_to_length=self.audio_signal.signal_length)
+            cur_signal = nussl.audio_signal.AudioSignal(audio_data_array=source_estimate,
+                                                        sample_rate=self.sample_rate)
+            signals.append(cur_signal)
+        return signals
+
+    def plot(self, output_name, three_d_plot=False, normalize=True):
+        """Plots histograms with the results of the DUET algorithm
```
ethman: Did you test this?

nussl/separation/duet.py
```diff
@@ -0,0 +1,440 @@
...
+        return smooted_matrix
+
+    def make_audio_signals(self):
+        """Returns the extracted signals
+
+        Returns:
+            signals (List[AudioSignal]): List of AudioSignals extracted using DUET.
+
+        """
+        signals = []
```
ethman: Because this is a public function, check to make sure that everything you need is set (correctly) before you do anything. So that's probably just making sure that all of the attributes are not `None`. You can probably just get away with checking to make sure that `len(self.masks) > 0` because the only way that the mask array is filled is if you ran the algorithm. If it fails, throw an exception.

nussl/separation/duet.py
```diff
@@ -0,0 +1,440 @@
...
+
+    def plot(self, output_name, three_d_plot=False, normalize=True):
+        """Plots histograms with the results of the DUET algorithm
+
+        Parameters:
+            output_name (str): path to save plot as
+            three_d_plot (Optional[bool]): Flags whether or not to plot in 3d. Defaults to False
+            normalize (Optional[bool]): Flags whether the matrix should be normalized or not
+        """
+        plt.close('all')
```
ethman: Same thing here as above, make sure that everything is set correctly before you do anything else.

nussl/separation/duet.py
```diff
@@ -0,0 +1,440 @@
...
+
+        aa = np.tile(self.attenuation_bins[1::], (self.num_delay_bins, 1)).T
```
ethman: `[1::]`! Double colon! AAAHHHH!!!!

nussl/separation/duet.py
```diff
@@ -0,0 +1,440 @@
...
+        dd = np.tile(self.delay_bins[1::].T, (self.num_attenuation_bins, 1))
```
ethman: AAAAAA**AAAAAAAA**aAAAaAAAaAaaoooaoaoaOAOOaoA_OAooAoOAOAOaA_**AAoooOOOoooMMMYYYYGGOOODDD**

tests/test_duet.py
```diff
@@ -0,0 +1,228 @@
...
+        duet.delay_peak = self.benchmark_dict['benchmark_delay_peak']
+        duet.atn_peak = self.benchmark_dict['benchmark_atn_peak']
+
+        mask_path = os.path.join('duet_reference', 'duet_benchmarks', 'benchmark_masks.npy')
+        benchmark_mask = np.load(mask_path)
+
+        masks = duet._compute_masks()
+        for i in range(len(masks)):
+            assert np.all(benchmark_mask[i].mask == masks[i].mask)
+
```
ethman: One more test case I want you to write: Instantiate a `Duet` object, run it, then run the same object again on a different audio signal. Make sure it doesn't crash and the audio sounds fine in both. (I've had some issues with running the same object on two different). Do it like this:
 `duet = nussl.Duet(signal1)`
`duet.run()`
`duet.audio_signal = signal2`
`duet.run()`
Etc, but fill in the blanks wrt setting up the input and output signals.

nussl/audio_signal.py
```diff
@@ -108,16 +109,34 @@ def __str__(self):
...
-    #     Returns:
     #
-    #     """
-    #     signal = self.audio_data
-    #     for channel in num_channels
 
+    def plot_time_domain(self, x_label='time', save=False, name=None, output_path=None):
+        """
+        Parameters:
+            x_label (str): Label the x axis with time or samples
```
ethman: If `x_label` can only be one of two values (time or samples) then it should be a bool. And in that case, maybe rename it to something more appropriate. Maybe `x_label_samples=False`, not sure though. Think of a better name...

nussl/audio_signal.py
```diff
@@ -108,16 +109,34 @@ def __str__(self):
...
+            save (bool): Save a png of the plot to the current folder
+            name (str): The name of the audio signal. Applies to title and filename
+            output_path (str): The output path of where the plot is saved if save is True
+
+        """
+        signal = self.audio_data
```
ethman: Don't copy this data if it's already available

nussl/audio_signal.py
```diff
@@ -108,16 +109,34 @@ def __str__(self):
...
+        for i in range(self.num_channels):
+            plt.subplot(self.num_channels+1, 1, i+1)
+            if x_label == 'time':
+                x_axis = np.linspace(0, len(signal[i])/self.sample_rate, len(signal[i]))
```
ethman: There is already a property that does exactly what this line does. Look at `self.time_vector`.

nussl/audio_signal.py
```diff
@@ -108,16 +109,34 @@ def __str__(self):
...
+                plt.plot(x_axis, signal[i])
+            else:
+                plt.plot(signal[i])
+            channel_num_plot = 'Channel ' + str(i)
+            plt.ylabel(channel_num_plot)
+        plot_title = name if name else 'Audio Signal'
```
ethman: Why not check if there is a `self.file_name`?

nussl/audio_signal.py
```diff
@@ -108,16 +109,34 @@ def __str__(self):
...
+        """
+        signal = self.audio_data
+        for i in range(self.num_channels):
+            plt.subplot(self.num_channels+1, 1, i+1)
+            if x_label == 'time':
+                x_axis = np.linspace(0, len(signal[i])/self.sample_rate, len(signal[i]))
+                plt.plot(x_axis, signal[i])
+            else:
+                plt.plot(signal[i])
+            channel_num_plot = 'Channel ' + str(i)
```
ethman: Use python's string format syntax here. That works like this: `'five plus three equals {}.'.format(5+3)` ==> `five plus three equals 8`

nussl/audio_signal.py
```diff
@@ -108,16 +109,34 @@ def __str__(self):
...
+            plt.ylabel(channel_num_plot)
+        plot_title = name if name else 'Audio Signal'
+        plt.suptitle(plot_title)
+        if save:
+            output_name_path = output_path if output_path else ''
+            output_name_path += name if name else 'audio_signal'
+            plt.savefig(output_name_path)
+        plt.show()
```
ethman: Don't do `plt.show()`

nussl/audio_signal.py
ethman: Also what if `self.audio_data` is None?

nussl/audio_signal.py
```diff
@@ -108,16 +109,34 @@ def __str__(self):
...
+                plt.plot(x_axis, signal[i])
+            else:
+                plt.plot(signal[i])
+            channel_num_plot = 'Channel ' + str(i)
+            plt.ylabel(channel_num_plot)
+        plot_title = name if name else 'Audio Signal'
+        plt.suptitle(plot_title)
+        if save:
+            output_name_path = output_path if output_path else ''
+            output_name_path += name if name else 'audio_signal'
```
ethman: Again, check `self.file_name` here.

nussl/audio_signal.py
```diff
@@ -108,16 +109,34 @@ def __str__(self):
...
-    # def plot_time_domain(self):
-    #     """
-    #     Not implemented yet -- will raise and exception
-    #     Returns:
     #
-    #     """
-    #     signal = self.audio_data
-    #     for channel in num_channels
 
+    def plot_time_domain(self, x_label='time', save=False, name=None, output_path=None):
```
ethman: Some questions: How do I plot only one channel? Can I make a stereo graph like librosa? What if I want to normalize the waveform (to [-1, 1])? Is the image size always the matplotlib default (i.e., you should change it to better aspect ratio if it needs it.

Look at some of the other plotting functions in `nussl`: Repet, RepetSim, AudioSignal.plot_spectrogram, Duet. Look at the logic, the options they have, how the manage user input, etc, and steal liberally.

PR #102, opened 2017-11-30
Adding support for downloading large files externally

ethman:
#### Reference Issue
Many files in the `nussl` github repo are very large for example, audio example files in .wav formats, test benchmark files in .mat or .npy binaries, and deep learning models. Many users do not need to download such large files when cloning the repo. It is therefore desirable to have utilities for the user to fetch these files from a remote server as needed. 

This pull request add utilities for downloading files from the `nussl-extras` repo.


ethman:
Deleting because this work was concurrently finished in the dev branch.

PR #120, opened 2018-09-22
Update melodia.py

### LINKED ISSUE #123: Onboarding users new to source separation

pseeth:


nathanshelly:
Removed all `git-lfs` files, do not reopen any PRs before this day.

PR #136, opened 2019-01-16
Reformatting nussl with black

### LINKED ISSUE #123: Onboarding users new to source separation

pseeth:
Ran [black](https://github.com/ambv/black) on nussl's train branch to enforce coding standards. Black was run with all defaults. Line length is 88.

Running [black](https://github.com/ambv/black) should have a number of benefits:

1. It enforces coding standards for all future contributors to nussl.
2. It's easier to read diffs with a standard line length of 88
3. While files did get longer, I think the code got easier to read.

#### Any other comments?
Also added in this is a pre-commit hook as suggested by the maintainers of black: https://black.readthedocs.io/en/stable/version_control_integration.html


nathanshelly:
Looks good to me! Would be nice to set up working CI builds at some point but as far as I can tell them failing isn't related to this commit, right?

pseeth:
CI builds just aren't working right now. We should get those up and running soon, I suppose.

pseeth:
Some odd things happening with comments trailing statements. Related to this black issue: https://github.com/ambv/black/issues/374.

ethman:
Yeah wrt the trailing comments, we should move them to before the relevant lines of code. Those are important for generating documentation. As long as this isn't being merged into master as is, it's not a pressing issue. But before pushing `train` to `master` (and then `docs`) this should be fixed.

pseeth:
Sounds good. I'm gonna go ahead and merge it into train then. Hopefully black gets to fixing that stuff soon.

ethman:
Can you make the f-strings all like `"this is a {}'.format('string')`. I don't think the new option is supported in older versions of python4/

ethman:
I didn't finish my review yet @pseeth lol

pseeth:
Should be supported completely in Python3.6 and above. They make string formatting a lot cleaner 
 so I'd like to keep them...who is on Python 3.5 and below anyway? And whoops! My bad...just keep commenting I suppose. I'll fix things in train before we merge to master.

ethman:
Well, I know gordon is still using python 3.5. I just don't want to alienate people before we have to. A lot of people aren't so quick to adopt.

PR #138, opened 2019-01-16
Black (try #2)

### LINKED ISSUE #123: Onboarding users new to source separation

pseeth:
Ran [black](https://github.com/ambv/black) on nussl's train branch to enforce coding standards. Black was run with all defaults. Line length is 88.

Running [black](https://github.com/ambv/black) should have a number of benefits:

1. It enforces coding standards for all future contributors to nussl.
2. It's easier to read diffs with a standard line length of 88
3. While files did get longer, I think the code got easier to read.

#### Any other comments?
Also added in this is a pre-commit hook as suggested by the maintainers of black: https://black.readthedocs.io/en/stable/version_control_integration.html


nussl/core/constants.py
```diff
@@ -54,9 +54,8 @@
 """list(str): list of all available windows in *nussl*
 """
 
-NUMPY_JSON_KEY = (
-    "py/numpy.ndarray"
-)  #: (str): key used when turning numpy arrays into json
+NUMPY_JSON_KEY = "py/numpy.ndarray"
```
nathanshelly: Super minor but could we put the comment above the constant? I think the way it is currently it's a little less clear which comments refer to which lines of code.

ethman: I don't recall the details 100%, but I remember that I put the comments after the constant because that's how the sphinx formatting worked to pick up single line comments. We might be able to finagle something better though...

ethman:
Just a general request: can we pause this black work until we agree on a formatting process? The formatting is super non-critical and I don't really have the bandwidth to look into other options. I see no reason why this can't wait until I'm back in a few months and we can properly evaluate what else is out there and come to a compromise. I can't say I'm happy with how this is being steamrolled through, despite my clear protestations.

PR #152, opened 2019-02-22
Fix potential bug in mix_source_folder_dataset

### LINKED ISSUE #123: Onboarding users new to source separation

ericye16:
#### Reference Issue

Very small, hopefully self-evident fix. I can create an issue for tracking if you'd like but hopefully that is not necessary.


pseeth:
Whoops, this must have slipped in from a renaming or something. Forgot the last bit! Definitely looks like a bug, merging your PR in now. Thanks!

PR #153, opened 2019-02-25
Add flag to resume training

### LINKED ISSUE #123: Onboarding users new to source separation

ericye16:
Adds the ability to resume training from the latest checkpoint.

#### Any other comments?
This is just a change I made in my branch which I thought might be appreciated upstream.

I'm not sure whether my changes to `resume()` are correct but they seem to work (it seems like there's a mismatch between what it expected to be saved and what was).

Also, I don't think the epoch number is handled correctly so the tensorboard might be overwritten with the new epoch numbers, but I don't know how to fix that.

### DIFF
--- a/bin/nussl-train
+++ b/bin/nussl-train
@@ -40,11 +40,18 @@ def parse():
         type=bool,
         help='Whether to log loss and other metrics to tensorboard file, defaults to True',
     )
+
+    parser.add_argument(
+        '--resume',
+        default=False,
+        type=bool,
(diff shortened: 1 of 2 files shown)


pseeth:
Hi there, thanks so much for the PR and really sorry for the late response. All the training stuff in nussl has been refactored to use PyTorch Ignite. So you can write resume functionality outside of nussl using PyTorch Ignite functionality instead.

PR #162, opened 2020-05-08
Adding augmentation utilities for AudioSignal objects

abugler:
This pull request adds `augmentation.py`, which includes augmentation utilities for AudioSignal objects, using ffmpeg. This includes:
 + Tremolo
 + Vibrato
 + Low/High pass
 + Chorus
 + Flanger
 + Phaser
 + Emphasis
 + Compression
 + Equalization
 + Pitch shifting
 + Time Stretching


nussl/core/augmentation.py
```diff
@@ -0,0 +1,377 @@
...
+import os
+from .audio_signal import AudioSignal
+import tempfile
+from .augmentation_utils import *
+
+# These values are found in the ffmpeg documentation in filters
+# that use the level_in arugment, however, not all filters that
+# use this state these bounds in documentationwith pytest.raises(ValueError):
+LEVEL_MIN = .015625
+LEVEL_MAX = 64
```
pseeth, line 13: These should be in `constants.py`.

nussl/core/augmentation.py
```diff
@@ -0,0 +1,377 @@
...
+
+    for audio_row in audio_signal.get_channels():
+        stretched_audio_data.append(librosa.effects.time_stretch(audio_row, stretch_factor))
+    stretched_audio_data = np.array(stretched_audio_data)
+
+    # The following line causes a UserWarning.
+    # stretched_signal = audio_signal.make_copy_with_audio_data(stretched_audio_data)
+
+    # This one does not
+    stretched_signal = AudioSignal(audio_data_array=stretched_audio_data)
```
pseeth, line 39: I think we should use the one that throws a warning. It's more clear. We can decide whether to filter the warning later. But right now, the STFT parameters will not be shared if you do it this way.

nussl/core/augmentation.py
```diff
@@ -0,0 +1,377 @@
...
+    stretched_signal.stft()
```
pseeth, line 40: Why do we need to take the STFT here?

nussl/core/augmentation.py
```diff
@@ -0,0 +1,377 @@
...
+
+    sample_rate = audio_signal.sample_rate
+    shifted_audio_data = []
+    audio_data = audio_signal.audio_data
+
+    for audio_row in audio_signal.get_channels():
+        shifted_audio_data.append(librosa.effects.pitch_shift(audio_row, sample_rate, shift))
+    shifted_audio_data = np.array(shifted_audio_data)
+    shifted_signal = audio_signal.make_copy_with_audio_data(shifted_audio_data)
+    shifted_signal.stft()
```
pseeth, line 65: Like before, why take the STFT?

nussl/core/augmentation.py
```diff
@@ -0,0 +1,377 @@
...
+    """
+    Applies high pass filter
+    Args: 
+        audio_signal: An AudioSignal object
+        highest_freq: Threshold for high pass. Should be positive scalar
+    Returns:
+        augmented_signal: A copy of the original audio signal, with augmentations applied 
+    """
+    if not np.issubdtype(type(lowest_freq), np.number) or lowest_freq <= 0:
+        raise ValueError("lowest_freq should be positve scalar")
```
pseeth, line 100: positve -> positive

nussl/core/augmentation.py
```diff
@@ -0,0 +1,377 @@
...
+    Returns:
+        augmented_signal: A copy of the original audio signal, with augmentations applied 
+    """
+    if not np.issubdtype(type(mod_freq), np.number) or mod_freq < 0:
+        raise ValueError("mod_freq should be positve scalar")
+    
+    if not np.issubdtype(type(mod_depth), np.number) or mod_depth < 0 or mod_depth > 1:
+        raise ValueError("mod_depth should be positve scalar between 0 and 1.")
+
+    filter_kwargs = {
```
pseeth, line 129: Why not do this directly? Why create a kwargs dict first then use it?

nussl/core/augmentation.py
```diff
@@ -0,0 +1,377 @@
...
```
pseeth, line 154: kwargs dict?

nussl/core/augmentation.py
```diff
@@ -0,0 +1,377 @@
...
+    Applies Chorus Filter
+    Args:
+        audio_signal: An AudioSignal object
+        TODO: Write this documentation.
+    Returns:
+        augmented_signal: A copy of the original audio signal, with augmentations applied
+    """
+
+    # TODO: Arg checks
+    filter_kwargs = {
```
pseeth, line 175: kwargs dict?

nussl/core/augmentation.py
```diff
@@ -0,0 +1,377 @@
...
+        "in_gain": in_gain,
+        "out_gain": out_gain,
+        "delays": make_arglist_ffmpeg(delays),
```
pseeth, line 178: I think these utilities should go into this folder, instead of into `augmentation_utils.py`.

nussl/core/augmentation.py
```diff
@@ -0,0 +1,377 @@
...
+
+    Applies Phaser Filter
+    Args:
+        audio_signal: An AudioSignal object
+        TODO: Write this documentation.
+    Returns:
+        augmented_signal: A copy of the original audio signal, with augmentations applied
+    """
+    # TODO: Arg checks
+    filter_kwargs = {
```
pseeth, line 199: kwargs dict

nussl/core/augmentation.py
```diff
@@ -0,0 +1,377 @@
...
+    https://ffmpeg.org/ffmpeg-all.html#flanger
+
+    Args:
+        audio_signal: An AudioSignal object
+        TODO: Write this documentation.
+    Returns:
+        augmented_signal: A copy of the original audio signal, with augmentations applied
+    """
+    # TODO: Arg checks
+    filter_kwargs = {
```
pseeth, line 222: kwargs dict

nussl/core/augmentation.py
```diff
@@ -0,0 +1,377 @@
...
+        level_in: Input gain
+        level_out: Output gain
+        _type: medium type to convert/deconvert from 
+        mode: reproduction to convert from physical medium, 
+            production to convert to physical medium.
+
+    Returns:
+        augmented_signal: A copy of the original audio signal, with augmentations applied 
+    """
+    if level_in < LEVEL_MIN or level_in > LEVEL_MAX \
```
pseeth, line 252: `constants.LEVEL_MIN/LEVEL_MAX` once you make the changes.

nussl/core/augmentation.py
```diff
@@ -0,0 +1,377 @@
...
+        raise ValueError("mod_depth should be positve scalar between 0 and 1.")
+
+    filter_kwargs = {
+        "f": mod_freq,
+        "d": mod_depth
+    }
+
+    return apply_ffmpeg_filter(audio_signal, "vibrato", **filter_kwargs)
+
+def chorus(audio_signal, in_gain=.4, out_gain=.4, delays=[50],
```
ethman, line 161:
```suggestion
def chorus(audio_signal, in_gain=.4, out_gain=.4, delays=[50], 
decays=[.8], speeds=[.95], depths=[.7]):
```
Don't make default arguments mutable: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

nussl/core/augmentation.py
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    return apply_ffmpeg_filter(audio_signal, "aemphasis", **filter_kwargs)
+
+def _compressor_argcheck(level_in, mode, reduction_ratio,
+    attack, release, makeup, knee, link,
+    detection, mix, threshold):
+
+    # The following values are taken from ffmpeg documentation
+    if level_in < LEVEL_MIN or level_in > LEVEL_MAX or \
```
pseeth, line 270: Use something more like

```
if (
   level_in < LEVEL_MIN or level_in > LEVEL_MAX or
   mode not in {"upward", "downward"} or 
)
```

Parens instead of \ for line breaks.

nussl/core/augmentation.py
```diff
@@ -0,0 +1,377 @@
...
+            'w': Width of the band in Hz
+            'g': Band gain in dB
+            't': Set filter type for band, optional, can be:
+                0, for Butterworth
+                1, for Chebyshev type 1
+                2, for Chebyshev type 2
+    """
+    # TODO: Argcheck
+
+    filter_kwargs = {
```
pseeth, line 363: I can see why you would use kwargs here.

nussl/core/augmentation.py
abugler: So for every ffmpeg filter, I found myself writing the code found in lines 52-71 in `augmentation_utils.py` in every single function. The only difference in each filter is the kwargs for `ffmpeg.filter()`. So this became the primary input to `apply_ffmpeg_filter`. If I was not to use kwargs here, I would have to declare every single possible filter argument in `apply_ffmpeg_filter`.

nussl/core/augmentation_utils.py
```diff
@@ -0,0 +1,72 @@
+import tempfile
+from .audio_signal import AudioSignal
+import ffmpeg
+
+def save_audio_signal_to_tempfile(audio_signal, extension=".wav"):
```
pseeth, line 5: We should use a different style for temporary files. See here for an example:

https://github.com/nussl/nussl/blob/master/nussl/core/play_utils.py#L147

Essentially, use a context manager to make sure everything is safe. This solution is also portable with Windows.

nussl/core/augmentation_utils.py
```diff
@@ -0,0 +1,72 @@
...
+
+def make_arglist_ffmpeg(lst):
+    return "|".join([str(s) for s in lst])
+
+# Passing this to an 'ffmpeg.input' will supress all ffmpeg
+# output. For troubleshooting, remove this dict from the function args
+silent_kwargs = {'loglevel': 'quiet'}
+
+def apply_ffmpeg_filter(audio_signal, _filter, silent=True, **kwargs):
+    audio_tempfile, audio_tempfile_name = \
```
pseeth, line 53: Use the tempfiles solution from before. All of this can be replaced with a good context manager.

tests/core/test_augmentation.py
```diff
@@ -0,0 +1,231 @@
+import numpy as np
+import numpy.random as random
+from nussl.core.augmentation import *
```
pseeth, line 3: Try not to import like this. Import explicitly, not implicitly.

nussl/core/augmentation.py
```diff
@@ -0,0 +1,377 @@
+import numpy as np
+import numpy.random as random
+import librosa
+import os
+from .audio_signal import AudioSignal
+import tempfile
+from .augmentation_utils import *
```
pseeth, line 7: Import explicitly the functions you need. I think these utils should be moved into `augment.py` directly, actually.

nussl/core/augmentation.py
abugler: If you make a copy, do you not have to take the STFT? I thought that did not automatically populate?

nussl/core/augmentation.py
pseeth: Instead of making a dictionary, you can pass them as keyword arguments to `apply_ffmpeg_filter`, instead of making a dictionary and then passing the dictionary.

nussl/core/augmentation.py
pseeth: The user should take the STFT instead, otherwise we are doing some side computation that the user didn't request and may not need. Function should only do what is specified.

nussl/core/augmentation.py
pseeth: Simple fix here is to do `delays=None`, then take care of it in the body of the function `delays = delays if delays is not None else [50]`.

nussl/core/augmentation.py
abugler: Just made that change in my code. I didn't actually know you could do that.

nussl/core/augmentation.py
```diff
@@ -4,7 +4,7 @@
 import os
 from .audio_signal import AudioSignal
 import tempfile
-from .augmentation_utils import *
+from .augmentation_utils import apply_ffmpeg_filter
```
pseeth, line 7: Let's move this into augment.py instead of keeping it separate.

abugler:
Changed `effects.py` such that all SoX effects use PySoX 1.4.0

ethman:
One last comment about the docs: as the API is getting a little more stable, think about adding some example code for each effect. The easier ones can be made in-line (see below), but the more involved ones we should link to an examples page in the docs.

Eg time stretch is easy:

    >>> audio_signal.signal_duration
    3.0
    >>> audio_signal.time_stretch(0.5)
    >>> audio_signal.signal_duration
    1.5

abugler:
In case it wasn't obvious by the last pull request, things I've changed:
 - I put some example code in `AudioSignal.build_effects`. There is also example code and usage notes at the stop of `effects.py`, since I didn't really see it to be necessary to put nearly the same  example code on every effect. 
 - Added argument types to many of the functions. 
 - Improved some Error handling

abugler:
Between commits 86779b3 and d8417a4, 4 more errors occur in unrelated files. These being `tests/ml/train_trainer.py`, `tests/ml/test_overfit.py`, and `tests/ml/tests_closures.py`. Unsure why this is happening, and will be investigating.

pseeth:
lgtm. @ethman?

nussl/core/audio_signal.py
```diff
@@ -1644,6 +1649,311 @@ def embed_audio(self, ext='.mp3', display=True):
         from . import play_utils
         return play_utils.embed_audio(self, ext=ext, display=display)
 
+    ##################################
+    #          Effect Hooks          #
+    ##################################
+
+    def build_effects(self, reset=True, overwrite=False):
```
ethman, line 1656: Does this apply effects to my audio or is this a preprocessing step before I can apply effects? If the former, this method should be renamed to `apply_effects`, if the latter then the documentation needs to be clearer.

nussl/core/audio_signal.py
```diff
@@ -1644,6 +1649,311 @@ def embed_audio(self, ext='.mp3', display=True):
...
+        """
+        Builds all effects in the AudioSignal's effects chains using the audio data of the calling 
+        AudioSignal. If reset=True, the current effects chains will be deleted. If overwrite=True,
```
ethman, line 1659: What exactly does `reset=True` do? Does it clear the effects chain before the effects are applied if there are any left over from last time?

nussl/core/audio_signal.py
```diff
@@ -1644,6 +1649,311 @@ def embed_audio(self, ext='.mp3', display=True):
...
+        Args:
+            reset (bool): If True, clears out all effects in effect chains. Default=True
+            overwrite (bool): If True, overwrites existing audio_data in AudioSignal. Defult=False
+                Also clears out stft_data
+        Returns:
+            self or new_signal (AudioSignal): If overwrite=True, then returns initially AudioSignal 
+            with edited audio_data. Otherwise, returns a new AudioSignal new_signal.
+        """
+        if not self._ffmpeg_effects_chain and not self._sox_effects_chain:
+            raise RuntimeError("No effect hooks have been called on this AudioSignal")
```
ethman, line 1685: I don't think this should raise an error. If there are no effects, it should just `return self` without making a fuss.

nussl/core/audio_signal.py
```diff
@@ -1644,6 +1649,311 @@ def embed_audio(self, ext='.mp3', display=True):
...
+        AudioSignal. If reset=True, the current effects chains will be deleted. If overwrite=True, 
+        the calling AudioSignal's audio_data will be overwritten and the stft_data will be deleted. 
+
+        FFmpeg effects will be applied AFTER SoX effects. The effects time_stretch and pitch_shift 
+        are SoX effects. All others are FFmpeg effects.
+
+        If no effects have been added to the AudioSignal's effects chains, then build_effects 
+        will throw a RuntimeError. 
+
+        This will add a tremolo effect followed by a high pass effect to the AudioSignal's effects chains:
```
ethman, line 1668: I think you need a few more examples here. For instance, if you start a very easy one like  `time_stretch`, you can show the length of the signal to demonstrate how it should work. E.g.,

`>>> my_signal.duration`
`10.0`
`>>> my_signal.time_stretch(0.5)`
`>>> my_signal.build_effects(overwrite=True)`
`>>> my_signal.duration`
`5.0`

Then you can demonstrate daisy chaining and other advanced things.

nussl/core/audio_signal.py
```diff
@@ -1644,6 +1649,311 @@ def embed_audio(self, ext='.mp3', display=True):
...
+        if reset:
+            self.reset_effects_chain()
+        if overwrite:
+            self.audio_data = new_signal.audio_data
+            self.stft_data = None
+            return self
+
+        return new_signal
+        
+    def make_effect(self, effect, **params):
```
ethman, line 1702: What does this function do?

nussl/core/audio_signal.py
```diff
@@ -1644,6 +1649,311 @@ def embed_audio(self, ext='.mp3', display=True):
...
+        """
+        Calls another effect hook.
+        Args:
+            effect (str): Function name of desired effect hook of the AudioSignal
+            **params: Additional parameters for given effect.
```
ethman, line 1707: Change `**params` to `**kwargs`.

nussl/core/constants.py
```diff
@@ -95,3 +95,10 @@
     ('accompaniment', 'bass+drums+other'), 
     ('linear_mixture', 'vocals+bass+drums+other')
 ])
+
+##### augmentation.py
+# These values are found in the ffmpeg documentation in filters
+# that use the level_in arugment, however, not all filters that
+# use this state these bounds in documentationwith pytest.raises(ValueError):
+LEVEL_MIN = .015625
+LEVEL_MAX = 64
```
ethman, line 104: PEP8: needs a newline at end of the file

nussl/core/effects.py
```diff
@@ -0,0 +1,584 @@
...
+        stream = ffmpeg.input(curr_tempfile.name, **input_args)
+        for _filter in filters:
+            stream = _filter(stream)
+        (stream
+         .output(out_tempfile.name)
+         .overwrite_output()
+         .run()
+         )
+
+        augmented_signal = AudioSignal(path_to_input_file=out_tempfile.name)
```
ethman, line 99: This won't keep metadata from the input `audio_signal`. I'm pretty sure the only metadata that you'll want to keep is `audio_signal.label`. This will also set `augmented_signal`'s path to the tempfile name, instead of the path of the input audio signal. I'm torn about whether this is okay or not. What are your thoughts @pseeth ?

nussl/core/effects.py
```diff
@@ -0,0 +1,584 @@
...
+
+def _flanger_argcheck(delay, depth, regen, width,
+                      speed, phase, shape, interp):
+    allowed_shape_types = {"triangular", "sinusoidal"}
+    if shape not in allowed_shape_types:
+        raise ValueError(f"shape must be one of the following:\n{allowed_shape_types}")
+    allowed_interp_types = {"linear", "quadratic"}
+    if interp not in allowed_interp_types:
+        raise ValueError(f"interp must be one of the following:\n{allowed_interp_types}")
+    if (delay < 0 or delay > 30
```
ethman, line 365: Two things here:
1) I think you might be able to just do `if not (0 < delay < 30):`
2) I think it might be worth splitting each of these in its own if statement and throwing a separate exception for each. Yes it's ugly, but this whole section is already pretty hairy, so we might as well make things easier for the user by telling them exactly what's wrong.

nussl/core/effects.py
```diff
@@ -0,0 +1,584 @@
...
+    }
+
+    return FFmpegFilter("aemphasis", level_in=level_in, level_out=level_out, mode=mode, **type_kwarg)
+
+
+def _compressor_argcheck(level_in, mode, reduction_ratio,
+                         attack, release, makeup, knee, link,
+                         detection, mix, threshold):
+    # The following values are taken from ffmpeg documentation
+    if (level_in < LEVEL_MIN or level_in > LEVEL_MAX or
```
ethman, line 469: Same comment as the flanger argcheck function

nussl/core/effects.py
abugler: I'll copy over in the `audio_signal.label` to the new signal. I think it would make sense if the `augmented_signal`'s path is just set to `None`.

ethman: You should also copy over effects metadata (https://github.com/nussl/nussl/pull/162#commitcomment-39648466)

abugler:
Demo Notebook :) https://nbviewer.jupyter.org/github/abugler/nussl_effects_notebook/blob/master/effects_notebook.ipynb

nussl/core/audio_signal.py
abugler: Removed Error

nussl/core/audio_signal.py
abugler: Removes effects in effects chains after applying effects. Made this more clear in docs

nussl/core/audio_signal.py
abugler: This applies effects. Changed function name

nussl/core/audio_signal.py
abugler: Added Example in docs

nussl/core/audio_signal.py
abugler: This function allows for the user to call a certain effect by passing the name of the effect as a string. I found this function very convenient when trying to make OnTheFly datasets.

abugler: changed docs to make it more clear

nussl/core/audio_signal.py
abugler: Renamed variable

nussl/core/constants.py
abugler: Fixed

nussl/core/effects.py
abugler: Changed to print only values that are out of bounds

nussl/core/effects.py
abugler: Fixed

nussl/core/audio_signal.py
```diff
@@ -201,6 +201,7 @@ def __init__(self, path_to_input_file=None, audio_data_array=None, stft=None, la
         # Effects
         self._ffmpeg_effects_chain = []
         self._sox_effects_chain = []
+        self.effects_applied = []
```
ethman, line 204: Awesome thanks! This needs to documented now in the Attributes section (here: https://github.com/nussl/nussl/blob/ce3f8792dcfe10c28c2ba2b5a8c393580137adec/nussl/core/audio_signal.py#L159)

nussl/core/audio_signal.py
```diff
@@ -1653,73 +1654,99 @@ def embed_audio(self, ext='.mp3', display=True):
     #          Effect Hooks          #
     ##################################
 
-    def build_effects(self, reset=True, overwrite=False):
+    def apply_effects(self, reset=True, overwrite=False):
         """
-        Builds all effects in the AudioSignal's effects chains using the audio data of the calling 
-        AudioSignal. If reset=True, the current effects chains will be deleted. If overwrite=True, 
-        the calling AudioSignal's audio_data will be overwritten and the stft_data will be deleted. 
+        Applies all effects in the AudioSignal's effects chains using the audio data of the calling
```
ethman, line 1659: I think this needs a little more fleshing out here. There are a few concepts that are not explicitly stated that need to be. Here is the place for us to spell out how the whole effects process works, conceptually. Questions that need to be addressed here: What's an "effects chain"?  How do I put effects in the chain? How are the effects stored/how can I access it to see what's there? Does the order of elements in the chain matter? What steps do I need to take before they're applied? Say it with language first, then talk about the details (what `reset`/`overwrite` do, which are FFmpeg vs SoX, etc), then provide a few bits of example code.

Also, you'll need to add one line of documentation to every effect function saying something to the _effect_ ;-) of "this won't get applied until you call `apply_effects()` and the order matters. for more details see link:`apply_effects()`". It's fine to copy and paste the same text to every effect function both here and in effects.py

nussl/core/audio_signal.py
```diff
@@ -1653,73 +1654,99 @@ def embed_audio(self, ext='.mp3', display=True):
...
         >>> audio_signal.tremolo(5, .6).high_pass(12000)
 
-        You can apply then apply the effects onto the AudioSignal with build_effects
+        Using overwrite here change the audio data of audio_signal, rather than create a new signal. 
+
+        >>> audio_signal.apply_effects(overwrite=True)
 
-        >>> audio_signal.build_effects(overwrite=True)
+        All effects applied will stored in `self.effects_applied`, a list of FilterFunctions
+        in the order that they were applied.
```
ethman, line 1695: I would also show the 1-liner version of this as icing on the cake: `new_signal = audio_signal.time_stretch(0.5).tremolo(5, .6).high_pass(12000).apply_effects()`

nussl/core/audio_signal.py
```diff
@@ -1653,73 +1654,99 @@ def embed_audio(self, ext='.mp3', display=True):
...
         except Exception:
-            raise AudioSignalException(f"Effect {effect} not found with parameters {params}")
+            raise AudioSignalException(f"Effect {effect} not found with parameters {kwargs}")
         
         return self
 
     def reset_effects_chain(self):
         """
-        Clears effects chain of AudioSignal
+        Clears effects chains of AudioSignal
```
ethman, line 1749: Add a line like "for more information on applying effects in AudioSignal see link:`apply_effects()`

nussl/core/audio_signal.py
ethman: Also add a reference to the explaining docs in `apply_effects()` ("for more info see...")

ethman:
Awesome work! I added a few more comments. I think the final things for me are all docs related, as noted inline above, but I'll add a checklist here:

- [ ] Do a thorough documentation conceptually explaining the whole process in `apply_effects()`
- [ ] Document `self.effects_applied`
- [ ] Add your notebook to the tutorials in the docs
- [ ] Make sure links work, both to other parts of the docs and to external pages (ffmpeg docs links)

See here for instructions on updating docs: https://nussl.github.io/docs/contributing.html#documentation

I talked to @pseeth  and we're on the same page about all of this.

pseeth:
Awesome work. @ethman, this looks good to me to merge? After merge, I'll update PyPI and update the docs page.

pseeth:
Just fixing up some docs building stuff. The chorus effect sounds a bit weird still on my end. If you can fix that, we're good to go, as long as tests pass. I also added the ability to pass in arbitrary arguments to both `time_stretch` and `pitch_shift`.

ethman:
I was looking over these changes again and I was wondering why SoX effects have to get applied before ffmpeg effects. Is there a major reason why this is? Is it _that_ much slower to apply the effects one-by-one (perhaps there's a way to batch while preserving order, e.g. if there are 6 ffmpegs in a row then a SoX)? I think, above all, we should honor the user's ordering.

ethman:
In `AudioSignal.applied_effects` can we also store the effect kwargs? Shouldn't be too hard, just overload a `FilterFunction.__str__()`. You'll have to migrate `**filter_kwargs` to that class and store them in `self`. Seems like there might be some other shared functionality you can push to that superclass as well.

abugler:
> I was looking over these changes again and I was wondering why SoX effects have to get applied before ffmpeg effects. Is there a major reason why this is? Is it _that_ much slower to apply the effects one-by-one (perhaps there's a way to batch while preserving order, e.g. if there are 6 ffmpegs in a row then a SoX)? Do you have benchmarks for "SoX-then-ffmpeg" vs. "user order"?
> 
> I think, above all, we should honor the user's ordering. Is there an engineering reason that would make this difficult?

This is something that Prem and I decided on quite awhile ago. The issue with FFmpeg effects, is that the audio data must be stored in a tempfile before processing. Therefore, it would much more efficient to store the signal in a tempfile, then apply several ffmpeg effects, then read the tempfile containing the signal. 

If all SoX effects came after FFmpeg effects, (or vice versa) it wouldn't impact the speed too much. However, if the user wanted to do effects in the order FFmpeg-SoX-FFmpeg, it would present more of an issue, as it would require two tempfile reads. 

I don't think it would be too much of an issue to implement user ordering, but speed is something that we were trying to prioritize at the time. @pseeth what do you think about this?

ethman:
> > I was looking over these changes again and I was wondering why SoX effects have to get applied before ffmpeg effects. Is there a major reason why this is? Is it _that_ much slower to apply the effects one-by-one (perhaps there's a way to batch while preserving order, e.g. if there are 6 ffmpegs in a row then a SoX)? Do you have benchmarks for "SoX-then-ffmpeg" vs. "user order"?
> > I think, above all, we should honor the user's ordering. Is there an engineering reason that would make this difficult?
> 
> This is something that Prem and I decided on quite awhile ago. The issue with FFmpeg effects, is that the audio data must be stored in a tempfile before processing. Therefore, it would much more efficient to store the signal in a tempfile, then apply several ffmpeg effects, then read the tempfile containing the signal.
> 
> If all SoX effects came after FFmpeg effects, (or vice versa) it wouldn't impact the speed too much. However, if the user wanted to do effects in the order FFmpeg-SoX-FFmpeg, it would present more of an issue, as it would require two tempfile reads.
> 
> I don't think it would be too much of an issue to implement user ordering, but speed is something that we were trying to prioritize at the time. @pseeth what do you think about this?

Do you have any side-by-side speed tests? I think that would answer whether or not we should do it.

abugler:
@ethman Just ran some speed tests. 
image

abugler:
With 1000 samples
image

ethman:
Okay, I just pushed some changes to your version, Andreas. Mostly touching up the docs, but you also missed propagating a handful of parameters which I fixed. 

At this point, my only remaining issue is with the implementation `_apply_user_ordered_effects()`. This is quite complicated and I cannot figure out how this is supposed to work in case it breaks in the future. I think a simpler and more interpretable solution might be to keep two indices around that define the start and end of the effects in the chain that are currently queued up for processing. Something like this:

lead_idx = 0
follow_idx = 0
while lead_idx <= len(fx_chain):
   if fx_chain[lead_idx] == fx_chain[follow_idx] or lead_idx == len(fx_chain):
      # same fx source, keep advancing the leader
      lead_idx += 1 
   else:
     # new fx source, apply effects and set follower to leader

abugler:
Yeah, I admit that function is very hacky. I'll try to clean it as much as possible!

ethman:
Hey Andreas, can I ask you to add a helper function to convert the `AudioSignal.effects_applied` list of strings to something we can use to reproduce the signal chain? So something along the lines of `[{'effect_name': 'fx1', 'effect_kwargs': {'param1': 0.1, ...}}, ...]`, that  parses the output of `FilterFunction.__str__()`. What I want is a list of things that we can pass to `AudioSignal.make_effect()` in a loop to completely apply the same effects more than once. You can put it in effects.py. Can you also write a test for the helper function?

Other than that, can you make `test_order()` more robust? Like make sure that the effects chain is applied in the way that we expect rather than just checking if the audio is not the same.

Sorry for this ask, I was going to do it myself but I won't have the bandwidth.

abugler:
No worries, I'll figure that out ASAP!

abugler:
> Hey Andreas, can I ask you to add a helper function to convert the `AudioSignal.effects_applied` list of strings to something we can use to reproduce the signal chain? So something along the lines of `[{'effect_name': 'fx1', 'effect_kwargs': {'param1': 0.1, ...}}, ...]`, that parses the output of `FilterFunction.__str__()`. What I want is a list of things that we can pass to `AudioSignal.make_effect()` in a loop to completely apply the same effects more than once. You can put it in effects.py. Can you also write a test for the helper function?
> 
> Other than that, can you make `test_order()` more robust? Like make sure that the effects chain is applied in the way that we expect rather than just checking if the audio is not the same.
> 
> Sorry for this ask, I was going to do it myself but I won't have the bandwidth.

I think it would actually be easier instead to store the `FilterFunction` objects as a list in `AudioSignal.effects_applied`.  Then instead, make `self._effects_chain` into a public variable, so you could do something like this:

```
signal2.effects_chain = signal1.effects_applied
```

Or for iteration:
```
for effect in signal1.effects_applied:
    filter_ = effect.filter
    params = effect.params
    signal2.make_effect(filter_, **params)
```

ethman:
Yeah, I think that's a great idea! Keep the `FilterFunction.__str__()` around because I'm sure that'll be useful. And also be sure to update AudioSignal constructor docs and add a sentence in `apply_effects()` about what's in the `effects_applied` list and how to use it (include your code snippet from your comment above).

Great work!

abugler:
So here are what I believe to be my final changes:
 + `effects_applied` and `effects_chain` are now properties of AudioSignal, and when you do `signal.effects_applied` or `signal.effects_chain`, it will result a **copy** of the desired list. This is to prevent madness like having two audio signals having the same list as an effects chain, where doing `signal.tremolo(.5, 6)` would add a tremolo effect to both signals. Sadly, it means this is not a valid operation:
```
signal2.effects_chain = signal1.effects_applied
```
 + `effects_applied` and `effects_chain` are both lists of `FilterFunction` objects. Since I did not originally plan on FilterFunction's being a class that user interact with, I cleaned up how the parameters are handled and stored, so that `FilterFunction.params` returns something helpful. Many input variables in FFmpeg are single letters, and that is originally how parameters where stored, so this has been changed too. 
+ You can now do this:
```
for effect in signal1.effects_applied:
    filter_ = effect.filter
    params = effect.params
    signal2.make_effect(filter_, **params)
```

These changes will be in the next commit :)

docs/changelog.rst
```diff
@@ -4,6 +4,7 @@ Changelog
 v1.1.1
 ------
 - Allowed MixSourceFolder to mix sources on the fly.
+- Added effects utilities.
```
pseeth, line 7: Make this more descriptive. Also add another line about fixing the bug in Ignite.

nussl/core/audio_signal.py
```diff
@@ -688,7 +697,8 @@ def log_magnitude_spectrogram_data(self):
         """
         (:obj:`np.ndarray`): Returns a real valued ``np.array`` with log magnitude spectrogram data.
         
-        The log magnitude spectrogram is defined as 20*log10(Abs(STFT)). Same shape as :attr:`stft_data`.
+        The log magnitude spectrogram is defined as 20*log10(Abs(STFT)).
```
pseeth, line 700: Just noticing this now, looks a bit odd. Change to `20 * log10(abs(stft))`.

nussl/core/audio_signal.py
```diff
@@ -705,6 +715,28 @@ def log_magnitude_spectrogram_data(self):
             raise AudioSignalException('Cannot calculate log_magnitude_spectrogram_data '
                                        'because self.stft_data is None')
         return 20 * np.log10(np.abs(self.stft_data) + 1e-8)
+    
+    @property
+    def effects_chain(self):
```
pseeth, line 720: Please add a note indicating how you CANNOT use these properties, per our offline discussion.

nussl/core/audio_signal.py
```diff
@@ -705,6 +715,28 @@ def log_magnitude_spectrogram_data(self):
...
+        (``list`` of ``nussl.core.FilterFunction``): Returns a copy of the AudioSignal's
+        effect chain. 
+
+        See Also:
+            * :func:`apply_effects`
+        """
+        return self._effects_chain.copy()
+
+    @property
+    def effects_applied(self):
```
pseeth, line 731: Add a note here as well, see comment on `effects_chain`.

nussl/core/constants.py
```diff
@@ -95,3 +96,10 @@
     ('accompaniment', 'bass+drums+other'), 
     ('linear_mixture', 'vocals+bass+drums+other')
 ])
+
+##### augmentation.py
+# These values are found in the ffmpeg documentation in filters
+# that use the level_in argument, however, not all filters that
+# use this state these bounds in documentation with pytest.raises(ValueError):
```
pseeth, line 103: Why is there a stray `pytest.raises` here? I don't fully understand this comment.

nussl/datasets/transforms.py
```diff
@@ -3,6 +3,7 @@
 import logging
 import random
 from collections import OrderedDict
+import librosa
```
pseeth, line 6: Why is there a new import here? I don't see where librosa is called later on. I think this is a stray import that you don't need anymore.

nussl/datasets/transforms.py
```diff
@@ -11,6 +12,7 @@
 from sklearn.preprocessing import OneHotEncoder
 
 from .. import utils
+from ..core import AudioSignal
```
pseeth, line 15: Don't think you need this import either.

requirements.txt
```diff
@@ -14,5 +14,9 @@ norbert
...
 termtables
 gpytorch
+ffmpeg-python
 importlib_resources
-pyloudnorm
\ No newline at end of file
+pyloudnorm
+sox==1.4.0b0
+numba==0.48
+soxbindings
```
pseeth, line 22: The `soxbindings` should move to a special way to install `nussl`: `pip install nussl[soxbindings]`, or `pip install -e ".[soxbindings]"` if installing from source. Do this in `setup.py`.

ethman:
Merging!

pseeth:
This is now in nussl 1.1.1. Thanks @abugler! Great work!

PR #164, opened 2020-07-28
Slakh dataset hook

abugler:
A hook for the slakh dataset. 
Includes submixing of the dataset with user-defined recipes.


abugler:
Replaced by #191

PR #169, opened 2020-08-07
Fixes a regularization error that caused exploding gradients in SI-SDR loss.

pseeth:
Also cleaned up one of the PIT test cases.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,9 +1,14 @@
 Changelog
 =========
 
+v1.1.2
+------
+- Fixed a bug in Dual Path models where they couldn't be used with DataParallel.
+
 v1.1.1
 ------
 - Allowed MixSourceFolder to mix sources on the fly.
+- Adding OnTheFly dataset for mixing sources on the fly using a function.
 - Added effects utilities for data augmentation.
 - Fixed bug with ignite 0.4.0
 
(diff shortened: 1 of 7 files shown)


ethman:
Can you explain the issue with the SI-SNR loss? Was the epsilon being too big really the problem? Also what's going on with the dims now?

pseeth:
The key was the `eps` in the `torch.log10` call, I believe. In any case it stabilized training. PIT also was reducing across both dimensions in the inner loop which was fine but unnecessary, due to the `torch.min` later in the code.

ethman:
Yeah, okay. Makes sense, I was confused by how the dimensions fit into the exploding gradient.

While you're in this loss function, can you add an `SNR_max` like in the MixIT papers? Scott Wisdom mentioned that clipping the gradient above some dB threshold prevented the net from ignoring poorer performing examples. Reference: https://openreview.net/pdf?id=qMMzJGRPT2d, and https://arxiv.org/pdf/2006.12701.pdf

pseeth:
The test appeared to be wrong (was comparing output of two different loss dictionaries that were not guaranteed to be the same). Fixed it - I think it was only incidentally passing before. The PIT loss change is definitely fine though.

ethman:
LGTM.

PR #174, opened 2020-08-18
Adding more metadata during training process.

ethman:
The goal of this PR is to add metadata to the SeparationModel object throughout its lifetime such that after it's trained and evaluated the metadata contains an exhaustive history of the model. The metadata will include the `config` (which contains architecture info), metadata on its training and validation sets, the loss over the training and validation sets, and evaluation datasets and metrics.

After the model goes through the whole training & evaluation pipeline, a user will be able to see and print this metadata as a record of how the model was made. This metadata will also be stored with the model on disk (in the same file as additional info passed to `torch.save()`) should the user choose to save their model.

This PR will make nussl models easier to store and manage in the long run. Additionally, this PR will streamline the process for adding models to the nussl EFZ by allowing us to just parse this metadata appropriately when we upload models.


ethman:
I haven't tested this at all or written any tests for this yet. I'm too tired right now, I'll double back soon. I'll ping you prem with this is ready to go.

ethman:
@pseeth, do you have a lot of models trained currently? It might be easier to break them and make a migration script than to support them going forward.

pseeth:
Can you describe the migration process? I can retrain more models but I can't speak for anyone that used this to train anything...

ethman:
The migration process would just make older models compatible with the new code, instead of supporting models from many different versions of nussl as it matures. Right now I've only rearranged the metadata that gets stored on disk when the model's saved, so this script would just reformat the old model's metadata so it matches the new format (with unknown fields left blank).

ethman:
And it would be part of nussl, so anyone could use it.

nussl/ml/train/closures.py
```diff
@@ -213,6 +214,13 @@ def __init__(self, loss_dictionary, optimizer, model, *args, **kwargs):
         self.optimizer = optimizer
         self.model = model
 
+        # Save about training metadata to model.info
+        self.model.metadata['optimizer'] = {
+            'name': type(optimizer).__name__,
+            'params': optimizer.defaults  # All of the settings are stored here.
```
pseeth, line 220: Does this only give defaults for the optimizer or does it give what was actually used during training?

nussl/separation/base/deep_mixin.py
```diff
@@ -27,20 +27,18 @@ def load_model(self, model_path, device='cpu'):
             metadata (dict): metadata associated with model, used for making
             the input data into the model.
         """
-        model_dict = torch.load(model_path, map_location='cpu')
-        model = SeparationModel(model_dict['config'])
-        model.load_state_dict(model_dict['state_dict'])
+        metadata = torch.load(model_path, map_location='cpu')
+        model = SeparationModel(metadata['config'])
```
pseeth, line 31: Hmm, not sure about the variable names here. `metadata` is the model? Can you add a description of `metadata` somewhere? I believe there's a section in the tutorial called "What's in a model".

tests/ml/test_overfit.py
```diff
@@ -84,7 +84,7 @@ def test_overfit_a(mix_source_folder):
             model_path, map_location=lambda storage, loc: storage)
         model.load_state_dict(state_dict['state_dict'])
 
-        history = state_dict['metadata']['trainer.state.epoch_history']
+        history = state_dict['trainer.state.epoch_history']
```
pseeth, line 87: These have moved to the top-level of the dictionary now?

nussl/ml/train/closures.py
ethman: Yeah, I tested it. It will reflect whatever values you pass into it. It's a confusingly named attribute, but I think the idea is that the params change throughout training so these are where they were initialized. ¯\_(ツ)_/¯

nussl/separation/base/deep_mixin.py
ethman: Still messing around with this a bit, it might change still.

tests/ml/test_overfit.py
ethman: Same as above.

ethman:
Still todo:

- [ ] Fix existing tests.
- [ ] Write migration tests and tests for new metadata.
- [ ] `model.print_metadata()` function.
- [ ] Figure out how to attach eval (meta)data to the model.
- [ ] Document.

pseeth:
This looks good to me - it's out of date with master at the moment though, so maybe update it to make sure tests pass?

ethman:
Yup, and I also want to capture our offline conversation on this, mainly that the last TODO is that the metadata should be attached to the model, not the separation object.

pseeth:
lgtm, let's wait for tests.

PR #175, opened 2020-08-25
Fixed sampling rate in sox effects issue, and tempo/stretch warning.

abugler:
### Preserved sampling rate in sox effects
Before, this would happen:
```python
# Assume new_sr is not 44100
>>> signal.resample(new_sr)
>>> new_signal = signal.pitch_shift(2).apply_effects()
>>> new_signal.sample_rate == new_sr
False
```
This is fixed in this pull request, and a new test case is added.

### Time stretch calls `sox.transforms.stretch` when factor is close to one.
Before, calling `time_stretch` with `0.9 <= factor <= 1.1` would cause this warning:
```python
"For this stretch factor, the stretch effect has better performance."
```
Now, if  `0.9 <= factor <= 1.1`, `stretch` is called instead of `tempo`.

References:
1. https://buildmedia.readthedocs.org/media/pdf/pysox/latest/pysox.pdf


ethman:
I'm not really sure what's happening here. Does sox actally call `time_stretch` when the user tries `pitch_shift` around 0.1? Does _that_ change the sample rate? Or is this an error on our end, where we're not calling things correctly so the sample rate isn't preserved?

ethman:
Also, can you point me to a specific page in the pysox PDF?

pseeth:
No, if you call `resample`, then call an effects chain, the sample rate reverts to `constants.DEFAULT_SAMPLE_RATE` because of this line we didn't catch:

https://github.com/nussl/nussl/blob/acb62b6cd29b5fdd08539bae905d2df1d83d62a0/nussl/core/effects.py#L163-L166

This doesn't happen for `ffmpeg` as those are written to disk with the new sample rate and read back into an `AudioSignal`. So this is a side effect introduced when we switched to `build_array`, which doesn't correct the sample rate. I believe these should be using something like `audio_signal.make_copy_with_audio_data(audio_data)` calls or something.

pseeth:
@ethman  Click through to source: https://pysox.readthedocs.io/en/latest/api.html#sox.transform.Transformer.tempo

pseeth:
I'm wondering if we should keep the warning but also add a stretch effect? Or should our `time_stretch` effect switch dynamically under the hood between `tempo` and `stretch` based on the factor?

abugler:
> I'm wondering if we should keep the warning but also add a stretch effect? Or should our `time_stretch` effect switch dynamically under the hood between `tempo` and `stretch` based on the factor?

I like the second option more, especially when PySox urges you to use tempo or stretch based on specific bounds.

ethman:
>> I'm wondering if we should keep the warning but also add a stretch effect? Or should our time_stretch effect switch dynamically under the hood between tempo and stretch based on the factor?

> I like the second option more, especially when PySox urges you to use tempo or stretch based on specific bounds.

I think if the output is the same when we dynamically switch, then I think that's the best option

pseeth:
Okay this sounds good but maybe we should at least propagate the warning up to the user?

abugler:
> > > I'm wondering if we should keep the warning but also add a stretch effect? Or should our time_stretch effect switch dynamically under the hood between tempo and stretch based on the factor?
> 
> > I like the second option more, especially when PySox urges you to use tempo or stretch based on specific bounds.
> 
> If the output is the same when we dynamically switch, then I think that's the best option

I did some testing, and I did find some differences between using `tempo` and using `stretch`. However, I don't think there will ever be a good reason why someone would use a less-optimal algorithm of time-stretching when a better one is available.

> Okay this sounds good but maybe we should at least propagate the warning up to the user?

If we are dynamically switching between `tempo` and `stretch`, we won't need to do this?

pseeth:
Can we split this into 2 PRs? The sample rate issue is a bug and the tempo/stretch issue can be done as a different PR, so I think beyond just being in the same file, they don't belong in the same PR.

abugler:
I'm splitting this into two PRs.

PR #177, opened 2020-08-27
Fixes to handlers and fixing some small bugs.

pseeth:
This PR fixes some odds and ends I ran into recently. I'll do a review of it myself inline. 

Summary of changes:

- Removed the output norm in DualPath
- Added a gated sigmoid activation function to Embedding
- Added some regularization of stuff.
- Fixed the tensorboard handler so it doesn't save a new file for every epoch.
- Fixed the time elapsed handler so it doesn't wrap around at a day.

### DIFF
--- a/nussl/ml/networks/modules/blocks.py
+++ b/nussl/ml/networks/modules/blocks.py
@@ -224,7 +224,7 @@ class LayerNorm(nn.Module):
             mean_dims.append(len(data.shape) - 1)
         mean = data.mean(dim=mean_dims, keepdims=True)
         var = data.var(dim=mean_dims, keepdims=True)
-        data = (data - mean) / ((var + 1e-5) ** 0.5)
+        data = (data - mean) / ((var + 1e-8) ** 0.5)
         data = data * self.gamma + self.beta
         # move everything back and squeeze off extra 
         # dimensions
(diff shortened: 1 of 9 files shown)


nussl/ml/networks/modules/blocks.py
```diff
@@ -324,7 +324,7 @@ class Embedding(nn.Module):
         embedding_size (int): Dimensionality of embedding.
         
         activation (list of str): Activation functions to be applied. Options 
-          are 'sigmoid', 'tanh', 'softmax', 'relu'. Unit normalization can be applied by 
+          are 'sigmoid', 'tanh', 'softmax', 'relu', 'gate'. Unit normalization can be applied by
```
ethman, line 327: I'd rename this to `gated_tanh` to be more descriptive.

ethman (approved):
LGTM

nussl/ml/networks/modules/blocks.py
pseeth: Done, let's just wait on tests.

pseeth:
Merging this into master after tests pass. I caught one more issue - if ffmpeg isn't installed we throw a warning, but warnings hasn't been imported yet in `__init__.py` sooo everything crashes. Classic.

PR #178, opened 2020-08-30
Bug fix: Sample rate in sox effects

abugler:
## Fixed a bug: Sample rate not being preserved when SoX effects were applied. 

Before, this would happen:
```python
# Assume new_sr is not 44100
>>> signal.resample(new_sr)
>>> new_signal = signal.pitch_shift(2).apply_effects()
>>> new_signal.sample_rate == new_sr
False
```
This is fixed in this pull request, and a new test case is added.

One of the issues from #175


nussl/core/effects.py
```diff
@@ -159,8 +159,9 @@ def apply_effects_sox(audio_signal, filters):
         input_array=np.transpose(audio_data), 
         sample_rate_in=audio_signal.sample_rate
     ) 
-    
-    augmented_signal = AudioSignal(audio_data_array=np.transpose(augmented_data))
+
+    augmented_signal = AudioSignal(audio_data_array=np.transpose(augmented_data),
```
pseeth, line 163: Is there a solution that uses `make_copy_with_audio_data` instead? If done this way then other properties like `stft_params` won't propagate either, I think.

ethman: Also, is the `np.transpose` redundant? We check to make sure the shape is correct whenever `AudioSignal.audio_data` is set, so you can clean up this line by removing the transpose.

pseeth: Oh yeah good point. Also Andreas use `x.T`, not `np.transpose`, I think. Just a style thing.

abugler: would we want to make a copy of the audio signal for ffmpeg effects as well? right now, the label nor stft is cascaded for ffmpeg

pseeth: I think that would be good, yes. Add some test cases for both as well - thanks!

pseeth:
This looks good to me, but there's 1) a conflict with master, and 2) the tests appear to have stalled out. Can you fix the conflict and commit an update? That should hopefully fix the tests too...

pseeth:
See #185.

PR #179, opened 2020-08-30
Automatically use sox.stretch when factor is close to 1. 

abugler:
### Time stretch calls `sox.transforms.stretch` when factor is close to one.
Before, calling `time_stretch` with `0.9 <= factor <= 1.1` would cause this warning:
```python
"For this stretch factor, the stretch effect has better performance."
```
Now, if  `0.9 <= factor <= 1.1`, `stretch` is called instead of `tempo`.

According to PySox documentation, `stretch` has much better performance than `tempo` when `factor is close to 1, so I don't see a reason to let the user choose the stretch algorithm with worse performance for a specific range.


pseeth:
Can you post some audio examples comparing the two somewhere?

abugler:
Here is a Jupyter Notebook containing the differences of using tempo vs. stretch: https://nbviewer.jupyter.org/github/abugler/sox_tempo_stretch/blob/master/Tempo_Stretch.ipynb

According to PySox Documentation, the factor parameter for both tempo and stretch should be "The ratio of the new tempo to the old tempo.". However, the stretch effect appears to be the other way around, while the tempo effect is working as intended. I think I found a PySox bug.

pseeth:
Let's merge master into this and then we can merge this.

pseeth:
Some care to be taken here - the regression data will conflict with master.

abugler:
Apparently `stretch` has the opposite effects than `tempo`, and the PySox maintainers will not change this anytime soon: https://github.com/rabitt/pysox/pull/116

I will change this such that, `tempo(.9 < x < 1.1)` will result in behavior if `transformer.tempo()` is called.

nussl/core/effects.py
```diff
@@ -129,7 +129,13 @@ class SoXFilter(FilterFunction):
     def __init__(self, filter_, **filter_kwargs):
         super().__init__(filter_, **filter_kwargs)
         if filter_ == "time_stretch":
-            self.func = lambda tfm: tfm.tempo(**filter_kwargs)
+            factor = filter_kwargs['factor']
+            if abs(factor - 1) <=  .1:
```
pseeth, line 133: Why not just `if .9 <= factor <= 1.1`?

Also, how were these bounds selected: something from PySox?

abugler: Yes, these are the bounds PySox use to determine the recommended algorithm.

abugler: `.9 <= factor <= 1.1` works too.

pseeth: I see now that your code matches what is in PySox:

https://github.com/rabitt/pysox/blob/master/sox/transform.py#L3192.

Okay cool, no change required then.

pseeth:
@abugler, please update the docstring for `time_stretch` when you get a chance to make it super clear that we are _dynamically_ choosing which type of stretching to apply! I also brought your PR up to date with master.

pseeth:
Can you also regenerate this notebook with your fixed code so we can compare the audio with stretch vs tempo?

https://nbviewer.jupyter.org/github/abugler/sox_tempo_stretch/blob/master/Tempo_Stretch.ipynb

abugler:
> Can you also regenerate this notebook with your fixed code so we can compare the audio with stretch vs tempo?
> 
> https://nbviewer.jupyter.org/github/abugler/sox_tempo_stretch/blob/master/Tempo_Stretch.ipynb

This doesn't use `nussl.core.effects`! It only uses nussl to embed the audio into the notebook.

pseeth:
> This doesn't use `nussl.core.effects`! It only uses nussl to embed the audio into the notebook.

Oh I see! Can you make it using `nussl` just so we can hear the difference without the inverse factor issue?

abugler:
I just fixed this notebook. (nbviewer won't render the new one)

pseeth:
Got it, so perhaps we won't merge this one then until we do some further investigation? Maybe we should raise this as an issue to the PySox devs? With audio examples, that is. 

It _might_ be better with speech signals vs music signals? Or with signals without drums vs with drums?

PR #184, opened 2020-09-23
Changes for passing filterbank tests and Ignite0.4.2 changes

abugler:
This PR does two things:
## Raises the atol for L118 for `tests/ml/test_filterbank.py` to `1e-5`
This is a small patch for now that should be fixed in a later update. @pseeth mentioned that the reason for this breaking is type changes. 
## `BaseDataset` now inherits `Iterable`
Pytorch-Ignite requires that any dataset used in a Engine inherits `Iterable`. I have also also implemented a simple `__iter__` function, as `Iterable` requires this.


ethman:
LGTM

nussl/datasets/base_dataset.py
```diff
@@ -227,6 +228,19 @@ def __getitem__(self, i):
 
         return data
 
+    def __iter__(self):
+        """
+        Calls ``self.__getitem__`` from ``0`` to ``self.__len__()``.
+        Required when inheiriting Iterable.
```
pseeth, line 234: Typo in docs. Also, I assume this function doesn't mess up anything else? Using DataLoader with `num_workers > 0` is okay?

pseeth (approved):
lgtm, just want to make sure that changing to an Iterable doesn't have any adverse effects.

pseeth:
LGTM, let's wait for tests and squash + merge.

nussl/datasets/base_dataset.py
abugler: I wrote a test, and it seems that this will work with num_workers>0. According to this thread: pytorch/pytorch#3636 (comment), I seem to be re-implementing what python would do we were just to iterate over an object with only __getitem__ implemented. (copied from deleted comment)

PR #186, opened 2020-10-01
Kwargs argument added for nussl.core.effects and effects hooks for audio signal.

abugler:
Added `**kwargs` argument to all effects functions in `nussl.core.effects` and all effects hooks in `nussl.core.audio_signal`. The `**kwargs`are passed to the respective SoX filter if the effect is a SoX effect, or `ffmpeg.filter` if the effect is an FFmpeg effect.


pseeth (approved):
lgtm

PR #187, opened 2020-10-03
Adds interaction via gradio.

pseeth:
This PR adds a pretty simple interaction to all nussl separation algorithms via gradio. By doing:

```python
separator.interact()
```

the following interface will be launched:

image

which allows you to upload audio and listen back to its separations. By doing:

```python
separator.interact(share=True)
```

a public link will be created that can be accessed by others. Pretty cool!

I added a simple smoke test for this to get coverage, and added it to the README. Here's a standalone demo script you can try out to see what this is like (be sure to `pip install gradio`) first:

```python
import nussl

# Make a dummy signal to instantiate object.
dummy_signal = nussl.AudioSignal()

# Create some separation object with desired config.
# This will use the 2DFT algorithm, but every method
# in nussl has interaction implemented.
ft2d = nussl.separation.primitive.FT2D(dummy_signal)

# Launch the Gradio interface. Use `share=True` to make
# a public link that you can share! Warning - people with
# access to the link will be running code on YOUR
# machine.
ft2d.interact()
```

`gradio` has a lot of random stuff in its dependencies so I only import it in the `interact` function. The core dependencies of nussl are not changed by this PR. Users can enable this by doing `pip install gradio`.


README.md
```diff
@@ -84,6 +84,37 @@ algorithms, including:
...
+
+```python
+import nussl
+
+# Make a dummy signal to instantiate object.
+dummy_signal = nussl.AudioSignal()
+
+# Create some separation object with desired config.
+# This will use the 2DFT algorithm, but every method
+# in nussl has interaction implemented.
```
ethman, line 101: How are these implemented? I assume they're all the same, right? I think this could be clearer.

Also, can you include an image of what it should look like here?

ethman:
Just a doc comment, other than that LGTM

README.md
pseeth: Did you one better, added a gif showing off the interface.

PR #188, opened 2020-10-05
Bug fix: safe model loading with transforms

pseeth:
Models that saved `transforms` to `metadata['transforms']` need the transforms moved to `metadata['train_dataset']['transforms']` in order to work and stay backwards compatible.

This PR fixes this bug, adds tests, and bumps to 1.1.3rc4.


ethman (approved):
LGTM

PR #189, opened 2020-10-07
Bug fix: module snapshot saving throws an error inside Jupyter notebook

pseeth:
If you define your modules in a Jupyter notebook, the snapshot errors out due to
this: https://bugs.python.org/issue12920. This PR fixes this by placing the
module snapshot code inside a try/catch, and puts a default string if the module snapshot could not be captured.


ethman (approved):
LGTM

Issue #172, opened 2020-08-11
Convolutional reverb


pseeth:
**Is your feature request related to a problem? Please describe.**
It'd be nice to add the ability to convolve signals with impulse responses, possibly multichannel impulse responses.

**Describe the solution you'd like**
I think we could implement it with the following API: 

```python
import nussl

audio = nussl.AudioSignal('/path/to/audio.wav')

# load impulse from file
impulse = nussl.AudioSignal('/path/to/impulse.wav')
# or load from array gotten by some means
impulse_array = generate_random_ir()
impulse = nussl.AudioSignal(audio_data_array=impulse_array, sample_rate=audio.sample_rate)

# non destructive (default), returns a new signal
convolved = audio.convolve(impulse) # -> AudioSignal

# destructive
audio.convolve(impulse, overwrite=True) # -> inplace
```

If `impulse` is single channel, then it is applied to all channels in `audio`. If `impulse` is multi-channel, then `audio` and `impulse` must have the same number of channels. Since `impulse` is an AudioSignal object, one could do `to_mono()` if there is a channel mismatch, then apply it:

```python
# fix for channel mismatch
convolved = audio.convolve(impulse.to_mono())
```

or use `make_audio_signal_from_channel`:

```python
convolved = audio.convolve(impulse.make_audio_signal_from_channel(0))
```

Treating impulse responses as just AudioSignal possibly gives a lot of nice benefits. Users could also convolve two signals even if one of them is not an impulse response, which could sound...interesting.

**Describe alternatives you've considered**
Implementing this outside is nussl is simple by just editing the audio_data arrays. But this could be a nice feature, especially for data augmentation.

**Additional context**

pseeth:
For actual implementation, we can just use `scipy.signal.convolve`, I think? We could do a GPU implementation via torch as well but I don't know if that's worth it.

ethman:
As per our offline discussions of this, I think this is a good proposal. It shouldn't be that hard to implement by adding our own logic around scipy's version, which is already a dependency.

pseeth:
Cool, I can get this going. @abugler, any thoughts?

pseeth:
One more note, I think this could go as a function that takes two AudioSignal objects in `mixing.py` perhaps? And the `AudioSignal.convolve` can call that function with `self` and `other`. Thoughts?

ethman:
[...]

PR #196, opened 2020-12-15
Adding convolution reverb function

pseeth:
This PR adds convolution reverb functionality to AudioSignal, following discussion in #172. The functionality is:

    Convolves signal one with signal two. There are three
    cases:
    
    1. s1 is multichannel and s2 is mono.
       -> s1's channels will all be convolved with s2.
    2. s1 is mono and s2 is multichannel.
       -> s1 will be convolved with each channel of s2.
    3. s1 and s2 are both multichannel.
       -> each channel will be convolved with the matching 
          channel. If they don't have the same number of
          channels, an error will be thrown.

### DIFF
--- a/nussl/core/audio_signal.py
+++ b/nussl/core/audio_signal.py
@@ -2330,6 +2330,64 @@ class AudioSignal(object):
         self._effects_chain.append(effects.equalizer(bands, **kwargs))
         return self
 
+    def convolve(self, other, method='auto', normalize=True, 
+             scale=True):
+        """
+        Convolves signal one with signal two. There are three
+        cases:
+        
+        1. s1 is multichannel and s2 is mono.
+        -> s1's channels will all be convolved with s2.
+        2. s1 is mono and s2 is multichannel.
+        -> s1 will be convolved with each channel of s2.
+        3. s1 and s2 are both multichannel.
+        -> each channel will be convolved with the matching 
+            channel. If they don't have the same number of
+            channels, an error will be thrown.
+
+        Args:
+            other: AudioSignal with which to convolve the two signals.
+            method: A string indicating which method to use to calculate the convolution. Options are:
+                direct: The convolution is determined directly from sums, the definition of convolution.
+                fft: The Fourier Transform is used to perform the convolution by calling fftconvolve.
+                auto: Automatically chooses direct or Fourier method based on an estimate of which is faster (default).
+            normalize: Whether to apply a normalization factor which will prevent clipping. Defaults to True.
+            scale: Whether to scale the output convolved signal to have the same max as the input, so they are of roughly equal loudness.
+        """
+        if other is None: return self
+        output = []
+        factor = 1.0
+        if normalize: factor = np.sum(np.abs(other.audio_data) ** 2)
+        
+        if self.num_channels != 1 and other.num_channels != 1:
+            if self.num_channels != other.num_channels:
+                raise AudioSignalException(
+                    "If both signals are multichannel, they must have the " 
+                    "same number of channels!")
+            for s1_ch, s2_ch in zip(self.get_channels(), other.get_channels()):
+                convolved_ch = scipy.signal.convolve(
+                    s1_ch, s2_ch / factor, mode='full', method=method)
+                output.append(convolved_ch)
+        else:
+            for i, s1_ch in enumerate(self.get_channels()):
+                for j, s2_ch in enumerate(other.get_channels()):
+                    convolved_ch = scipy.signal.convolve(
+                        s1_ch, s2_ch / factor, mode='full', method=method)
+                    output.append(convolved_ch)
+        
+        output = np.array(output)
+        if scale:
+            max_output = np.abs(output).max()
+            max_input = np.abs(self.audio_data).max()
+            scale_factor = max_input / max_output
+            output *= scale_factor
+        
+        convolved_signal = self.make_copy_with_audio_data(
+            output, verbose=False)
+        convolved_signal.truncate_samples(self.signal_length)
+                
+        return convolved_signal
+
     ##################################################
     #              Operator overloading              #
     ##################################################
--- a/tests/core/test_effects.py
+++ b/tests/core/test_effects.py
@@ -1,5 +1,6 @@
 from copy import deepcopy
 import numpy as np
+import nussl
 import nussl.core.effects as effects
 from nussl.core.audio_signal import AudioSignalException
 import os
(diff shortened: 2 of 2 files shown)


nussl/core/mixing.py
```diff
@@ -3,11 +3,129 @@
...
+        scale_factor = max_input / max_output
+        output *= scale_factor
+    
+    convolved_signal = signal.make_copy_with_audio_data(
+        output, verbose=False)
+    convolved_signal.truncate_samples(signal.signal_length)
+            
+    return convolved_signal
+
+MIN_LOUDNESS = -70
```
ethman, line 72: Should these go in nussl.constants?

pseeth: Moved em.

nussl/core/mixing.py
```diff
@@ -3,11 +3,139 @@
...
+        if signal.num_channels != other.num_channels:
+            raise RuntimeError(
+                "If both signals are multichannel, they must have the " 
+                "same number of channels!")
+        for s1_ch, s2_ch in zip(signal.get_channels(), other.get_channels()):
+            convolved_ch = scipy.signal.convolve(
+                s1_ch, s2_ch / factor, mode='full', method=method)
+            output.append(convolved_ch)
+    else:
+        for i, s1_ch in enumerate(signal.get_channels()):
```
ethman, line 66: Do you need these enumerates()? You're not using the index vars.

nussl/core/mixing.py
```diff
@@ -3,11 +3,139 @@
...
+    if signal.num_channels != 1 and other.num_channels != 1:
+        if signal.num_channels != other.num_channels:
+            raise RuntimeError(
+                "If both signals are multichannel, they must have the " 
+                "same number of channels!")
+        for s1_ch, s2_ch in zip(signal.get_channels(), other.get_channels()):
+            convolved_ch = scipy.signal.convolve(
+                s1_ch, s2_ch / factor, mode='full', method=method)
+            output.append(convolved_ch)
+    else:
```
ethman, line 65: Pedantic: Maybe add a comment saying that at least one of these for loops is over 1 item. Wasn't clear to me at first.

nussl/core/mixing.py
```diff
@@ -3,11 +3,139 @@
...
+    bg_signal = copy.deepcopy(bg_signal)
+
+    pad_len = max(0, fg_signal.signal_length - bg_signal.signal_length)
+    bg_signal.zero_pad(0, pad_len)
+    bg_signal.truncate_samples(fg_signal.signal_length)
+
+    n_loudness = max(MIN_LOUDNESS, bg_signal.loudness())
+    loudness = max(MIN_LOUDNESS, fg_signal.loudness())
+    
+    if loudness - snr < MIN_LOUDNESS:
```
ethman, line 108: Could this logic be replaced with `np.clip()`? If not, an explanatory comment would be nice to get an overview.

nussl/core/mixing.py
```diff
@@ -3,11 +3,139 @@
...
+        return_sources: Whether to return the sources or return the mix.
+    """
+    fg_signal = copy.deepcopy(fg_signal)
+    bg_signal = copy.deepcopy(bg_signal)
+
+    pad_len = max(0, fg_signal.signal_length - bg_signal.signal_length)
+    bg_signal.zero_pad(0, pad_len)
+    bg_signal.truncate_samples(fg_signal.signal_length)
+
+    n_loudness = max(MIN_LOUDNESS, bg_signal.loudness())
```
ethman, line 105: Can you find more descriptive variable names for `n_loudness`, `loudness`, and `t_loudness`? They're hard to keep straight @ line 126 when you do the computation.

nussl/core/mixing.py
```diff
@@ -3,11 +3,139 @@
...
+        inplace: Whether or not to do this operation in place.
+        return_sources: Whether to return the sources or return the mix.
+    """
+    fg_signal = copy.deepcopy(fg_signal)
+    bg_signal = copy.deepcopy(bg_signal)
+
+    pad_len = max(0, fg_signal.signal_length - bg_signal.signal_length)
+    bg_signal.zero_pad(0, pad_len)
+    bg_signal.truncate_samples(fg_signal.signal_length)
+
```
ethman, line 104: Seems like here we should be doing some checks to make sure things like sample_rate, n_channels, & length are OK. Maybe call `utils.verify_audio_signal_list_strict([fg_signal, bg_signal])` here.

nussl/core/mixing.py
```diff
@@ -3,11 +3,139 @@
...
+        fg_signal: AudioSignal object, will be the louder signal (or quieter if snr is negative).
+        bg_signal: AudioSignal object, will be the quieter signal (or louder if snr is negative).
+        snr: Signal-to-noise ratio in decibels to mix at.
+        inplace: Whether or not to do this operation in place.
+        return_sources: Whether to return the sources or return the mix.
+    """
+    fg_signal = copy.deepcopy(fg_signal)
+    bg_signal = copy.deepcopy(bg_signal)
+
+    pad_len = max(0, fg_signal.signal_length - bg_signal.signal_length)
```
ethman, line 101: Two things:

1. I think you should bubble up whether to do `min(len(fg), len(bg))` and truncate the longer one or `max(len(fg), len(bg))` and pad the shorter one. (I don't think the former is useful, but I figure why not give the option to the user?) I think the opinionated way you've written it now is too subtle and could cause unintended side effects.
2. Should that be it's own subroutine? Not sure if it's much use outside this function, but maybe `def _match_signal_lengths(signal_list, mode='pad')` or even something on AudioSignal like `signal1.match_length(signal2)`. Just spitballing, lmk what you think

nussl/core/audio_signal.py
```diff
@@ -2330,6 +2330,28 @@ def equalizer(self, bands, **kwargs):
...
+    ##################################################
+
+    def convolve(self, other, method='auto', normalize=True, 
+             scale=True):
+        from . import mixing
+        return mixing.convolve(self, other, method=method, 
+            normalize=normalize, scale=scale)
+
+    def mix(self, other, snr=10):
+        from . import mixing
```
ethman, line 2344: I'm assuming this lazy import is to avoid a circular import?

ethman (commented):
Overall looks good. Comments inline.

nussl/core/mixing.py
ethman: btw: By "bubble up" I mean add an arg for the user.

PR #203, opened 2021-02-15
SeparationModel improvements - load/save/flexible forward pass.

pseeth:
This PR improves SeparationModel a bit. Save and load functionality are now in SeparationModel. Keyword arguments can now be passed into a separation model like so:

```
separation_model(mix_audio=..., other_stuff=...)
```

Instead of needing to do

```
data = dict(mix_audio=..., other_stuff=...)
separation_model(data)
```

This change is backwards compatible, so the old way still works too.

### DIFF
--- a/nussl/ml/networks/separation_model.py
+++ b/nussl/ml/networks/separation_model.py
@@ -274,8 +274,10 @@ class SeparationModel(nn.Module):
             }
             metadata.update(dataset_metadata)
 
-        if val_data is not None:
+        try:
             metadata['val_dataset'] = _prep_metadata(val_data.metadata)
+        except:
+            pass
         
         if trainer is not None:
             train_metadata = {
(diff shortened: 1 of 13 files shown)


nussl/ml/networks/separation_model.py
```diff
@@ -8,7 +8,25 @@
...
+    if isinstance(transforms, datasets.transforms.Compose):
+        for t in transforms.transforms:
+            if isinstance(t, datasets.transforms.Cache):
+                transforms.transforms.remove(t)
+
+    return transforms
+
+
+def _prep_metadata(metadata):
+    metadata = copy.deepcopy(metadata)
```
ethman, line 26: Maybe a 1 line docstring for each of these funcs, now that their context has changed.

nussl/ml/networks/separation_model.py
```diff
@@ -190,10 +212,6 @@ def forward(self, data):
                     for k in stats[o]: 
                         stats_desc += f"\n\t\t\t{k}: {stats[o][k]:.4f}"
 
-                min_max = ", ".join(
```
ethman, line 193: Did you mean to remove this? What did this even do? Was this unused from a verbose mode or something?

ethman (approved):
LGTM.

PR #204, opened 2021-02-16
Changed FFmpeg effects in `nussl/core/effects.py` to save temporary audio file as `.wav`, rather than `.flac`.

abugler:
A PR to fix the flaky tests from  `tests/core/test_effects.py`, as a result of differences in reading .flac files between Github and Mystic. A much lower `atol` of 1e-4 is used instead of 1e-1 for effects. Some unrelated nussl tests needed the higher `atol`, so `atol` is now a kwarg of `check_against_regression_data`.


pseeth (commented):
lgtm

PR #205, opened 2021-02-16
Separation model updates

pseeth:
Update to separation model to fix a small issue when passing through arguments.


ethman:
What do these new changes solve? It's not obvious to me from the code.

pseeth:
Sorry. If you have a module that looks like this:

```python
class MyModule(nn.Module):
    def __init__(self):
        super().__init__()
    def forward(self, data=None, flip=False, **kwargs):
        if flip:
            data = -data
        return data
```

(taken from the test case). Then you would put it in the `config` so that it looked like this:

```
{
    "name": "Flipper",
    "modules": {
        "mix_magnitude": {},
        "test": {
            "class": "MyModule"
        }
    },
    "connections": [
        [
            "test",
            [
                "mix_magnitude",
                {
                    "embedding": "embedding",
                    "flip": false
                }
            ]
        ]
    ],
    "output": [
        "test"
    ]
}
```

If you wanted to change `flip` to `True`, it wouldn't work, as it never actually looked up the `key` in the data dictionary. So if you did

```python
model = SeparationModel(config)
model(data, flip=True) # would actually do flip=False, the default.
```

So this should now be fixed by looking up the key in the data, not just the value of the key (which was used previously for routing things around between blocks).

ethman (approved):
lgtm

PR #206, opened 2021-02-19
Adding overlap add functionality

pseeth:
This PR adds an algorithm that lets you do overlap add with any separation algorithm. The functionality is that the signal is split into overlapping windows. The windows overlap by half the window length. The algorithm is applied to each window in sequence. The windows are then recombined using overlap add with a Hanning window.

The test case is written using a dummy algorithm that does nothing to the signal but makes a copy of it. The signal coming out of overlap/add should be the same as the signal going into it, as the recombination shouldn't change anything. The test case just checks for this.


nussl/core/audio_signal.py
```diff
@@ -1222,6 +1223,7 @@ def truncate_seconds(self, n_seconds):
         """
```
ethman, line 1223: Update doc strings for all of these.

nussl/separation/composite/overlap_add.py
```diff
@@ -0,0 +1,89 @@
...
+from ... import AudioSignal
+import numpy as np
+import warnings
+import matplotlib.pyplot as plt
+
+class OverlapAdd(SeparationBase):
+    def __init__(self, separation_object, window_length=15, find_permutation=False):
+        """[summary]
+
+        Parameters
```
ethman, line 12: Wrong docstring style.

nussl/separation/composite/overlap_add.py
```diff
@@ -0,0 +1,89 @@
...
+            Separation object that overlap and add is applied to.
+        window_length : int, optional
+            Window length of overlap/add window, by default 15 seconds.
+            The hop length will be half the window length.
+        find_permutation : bool, optional
+            Whether or not to find the permutation between chunks before combining.
+        """
+        super().__init__(separation_object.audio_signal)
+
+        self.hop_length = window_length / 2
```
ethman, line 24: No way for user to set hop length?

nussl/separation/composite/overlap_add.py
```diff
@@ -0,0 +1,89 @@
...
+
+        Parameters
+        ----------
+        separation_object : SeparationBase
+            Separation object that overlap and add is applied to.
+        window_length : int, optional
+            Window length of overlap/add window, by default 15 seconds.
+            The hop length will be half the window length.
+        find_permutation : bool, optional
+            Whether or not to find the permutation between chunks before combining.
```
ethman, line 20: How is the permutation found?

nussl/separation/composite/overlap_add.py
```diff
@@ -0,0 +1,89 @@
...
+        audio_data = [audio_data for _ in range(len(estimates_from_chunks[0]['signals']))]
+        audio_data = np.stack(audio_data, axis=-1)
+
+        # Combine all the chunks, fixing permutations between them if needed        
+        for i, e in enumerate(estimates_from_chunks):
+            signals = e['signals']
+            if i > 0 and self.find_permutation:
+                # do something with permutation finding 
+                # (e.g. re-order signals in estimates['signals'], matching to previous set)
+                pass
```
ethman, line 71: oh. lol.

nussl/separation/composite/overlap_add.py
```diff
@@ -0,0 +1,89 @@
...
+        audio_signal.zero_pad(hop_samples, hop_samples)
+        num_samples = audio_signal.signal_length
+        
+        audio_data = np.zeros_like(audio_signal.audio_data)
+        chunk_starts = np.arange(0, num_samples - hop_samples, hop_samples)
+        estimates_from_chunks = []
+        
+        for i, start_idx in enumerate(chunk_starts):
+            end_idx = start_idx + win_samples
+            window = AudioSignal.get_window('hanning', win_samples)[None, :]
```
ethman, line 50: No way to set the window type?

nussl/separation/composite/overlap_add.py
```diff
@@ -0,0 +1,89 @@
...
+                'window': window,
+                'start_idx': start_idx,
+                'end_idx': end_idx
+            })
+
+        audio_data = [audio_data for _ in range(len(estimates_from_chunks[0]['signals']))]
+        audio_data = np.stack(audio_data, axis=-1)
+
+        # Combine all the chunks, fixing permutations between them if needed        
+        for i, e in enumerate(estimates_from_chunks):
```
ethman, line 66: Nit: this `e` variable name is a bit confusing later in the loop. Maybe `est`?

nussl/separation/composite/overlap_add.py
```diff
@@ -0,0 +1,89 @@
...
+        win_samples = int(self.window_length * audio_signal.sample_rate)
+        hop_samples = int(self.hop_length * audio_signal.sample_rate)
+        audio_signal.zero_pad(hop_samples, hop_samples)
+        num_samples = audio_signal.signal_length
+        
+        audio_data = np.zeros_like(audio_signal.audio_data)
+        chunk_starts = np.arange(0, num_samples - hop_samples, hop_samples)
+        estimates_from_chunks = []
+        
+        for i, start_idx in enumerate(chunk_starts):
```
ethman, line 48: One thing I wanted to do a while ago is make a generator attached to AudioSignal that would do all of this. So you would use it like:

```python
for region in audio_signal.get_next_region('hanning', win_samples):
    # do stuff
```

Something like that. It would do all of this fancy stuff you've done here within the AudioSignal object, so that the logic is usable elsewhere. Do you think that it would be worth moving what you have here? Would it be useful? Then this OverlapAdd isn't so much a class as it is a recipe. I haven't fully though through everything that would change. Thoughts?

nussl/separation/composite/overlap_add.py
pseeth: Only `hop_length = window_length / 2` will work, I think.

nussl/separation/composite/overlap_add.py
pseeth: Yeah will implement this soon. But will be using simple correlation of the output of each chunk.

nussl/separation/composite/overlap_add.py
pseeth: Hanning is the only perceptually valid one for overlap/add, here.

nussl/separation/composite/overlap_add.py
pseeth: Feels like an easy refactor in the future but wanted to keep this one simple and restricted to one method for now, if that's alright.

nussl/separation/composite/overlap_add.py
ethman: Yeah for COLA, but what if the user doesn't want COLA? What if they want to just apply a separator to non-overlapping chunks? That should be possible, no?

ethman (changes requested):
Overall lgtm, but I think needs more comments.

ethman (approved):
LGTM

PR #207, opened 2021-02-24
Making interaction more flexible.

pseeth:
Exposes/adds arguments to `separation_alg.interact()` that make it a bit more extensible. Backwards compatible with old calls to `interact`.

### DIFF
--- a/nussl/separation/base/separation_base.py
+++ b/nussl/separation/base/separation_base.py
@@ -78,7 +78,8 @@ class SeparationBase(object):
             else:
                 self._preprocess_audio_signal()
 
-    def interact(self, add_residual=False, source='upload', label=None, share=False):
+    def interact(self, add_residual=False, source='upload', label=None, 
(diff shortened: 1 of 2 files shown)


nussl/separation/base/separation_base.py
```diff
@@ -120,18 +124,20 @@ def _separate(file_obj): # pragma: no cover
                 estimates.append(mix - estimates[0])
 
             estimates = {f'Estimate {i}': s for i, s in enumerate(estimates)}
-            html = play_utils.multitrack(estimates, ext='.mp3', display=False)
+            html = play_utils.multitrack(estimates, ext=ext, display=False)
```
ethman, line 127: Maybe add the period if it's not there. `ext = f'.{ext}' if ext[0] != '.' else ext`

ethman (approved):

ethman:
lgtm

PR #209, opened 2021-02-28
Unpinning zarr + adding release.yml

pseeth:
Zarr has been pinned for 2.3.0 for some time. Let's try unpinning it and seeing if tests pass.


ethman (approved):
lgtm

PR #212, opened 2021-03-06
Small fix to expose overlap/add window type.

pseeth:
This PR exposes overlap and add's window_type to the user, setting it to default of 'hanning'. Not doing this was an oversight in the original OverlapAdd PR.


ethman:
lgtm. submit when tests pass. let me know if you need another CR

PR #213, opened 2021-03-06
torch.rfft -> torch.fft.rfft

pseeth:
PyTorch changed their FFT API to match numpy in PyTorch 1.8. This PR updates the one call we make to it.


ethman (approved):
lgtm

PR #228, opened 2021-06-26
Adding SalientExcerptMixSourceFolder object

### LINKED ISSUE #225: SalientExcerptMixSrc Dataset
Create a dataset that extends MixSrcFolder that only returns excerpt where a target source is salient.

ethman:
Fixes #225


nussl/datasets/hooks.py
```diff
@@ -417,6 +421,168 @@ def process_item(self, item):
...
+        )
+
+    def _populate_metadata(self):
+        """Makes a metadata dict of salient start times for each source."""
+
+        if self.verbose:
+            # TODO: logging.info()
+            print(f'Preprocessing data at {self.folder}.')
+
+        # TODO: Balanced dataset!
```
ethman, line 550: This is a feature we should talk about.

nussl/datasets/hooks.py
```diff
@@ -417,6 +421,168 @@ def process_item(self, item):
...
+                    'start': start
+                })
+        return metadata
+
+    def _get_mix(self, placeholder, item):
+        """OnTheFly closure. Gets called for every iteration to build a batch. A placeholder param is necessary since
+        OnTheFly will pass itself as well as the item to this function"""
+        item = self.song_metadata[item]
+        mixsrc_item = item['mixsrc_item']
+        # Note that the offset here is really the onset, but the parameter is passed as a kwarg to what is ultimately
```
ethman, line 579: Nit pick: Can this comment be more concise?

nussl/datasets/hooks.py
```diff
@@ -417,6 +421,168 @@ def process_item(self, item):
...
+        # the AudioSignal which takes the onset as the parameter "offset"
+        offset = item['start']
```
ethman, line 581: Did you test that this works with multiple threads? I.e., with a pytorch data loader with `n_workers > 1`? This seems like it could be a race condition, but I'm not 100% sure. If it is a race condition, then we need to rearchitect this object

nussl/datasets/hooks.py
```diff
@@ -187,22 +190,22 @@ def get_items(self, folder):
         ])
         return items
 
-    def get_mix_and_sources(self, item):
+    def get_mix_and_sources(self, item, **kwargs):
```
ethman, line 193: Why do these all need to be kwargs now?

ethman (commented):
Thanks for doing this! I've added inline comments, but I think the main takeaways for me are that I would expect the changes to be contained within the SalientExcerptMixSourceFolder object and related tests, but it seems like there's a bit of spillover. Overall, great work Boaz!

nussl/core/utils.py
```diff
@@ -211,6 +211,19 @@ def _format(string):
     return ''.join(list(filter(str.isalnum, string))).lower()
 
 
+def find_salient_starts(audio, duration_sec, hop_ratio, sr, threshold_db=-60.0):
+    """Finds frames in the audio where the RMS is above a dB threshold."""
+    dur = int(sr * duration_sec)
+    hop_dur = int(dur * hop_ratio)
+    threshold = np.power(10.0, threshold_db / 20.0)
+    # adjust the shape of a mono track
+    if audio.shape[0] == 1:
```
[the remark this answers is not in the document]
boazcogan: I'll double check, but I believe that the stereo worked fine, but for some reason mono needed to be formatted differently hence the special check.

nussl/datasets/hooks.py
```diff
@@ -417,6 +421,168 @@ def process_item(self, item):
                 "'mix', 'sources' as keys!")
         return output
 
+
+class SalientExcerptMixSourceFolder(OnTheFly):
+    # TODO: docs!
+    """
+    Given a dataset in the expected format, this class identifies and partitions the folder into a dataset containing
```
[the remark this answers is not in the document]
boazcogan: Got it I'll make that change.

tests/core/regression/multitrack_test.html
```diff
@@ -1,4 +1,4 @@
...
...Dl6+v/m7axTaVcRP79yQ+kpLSCWrp0tsHTwwl+p+amf7d2S1vK7N6YvrIdw0SmrNN20DmNTAUmvVZDZkyKIaAZvvfd3mMHZk0f/VGgv9fU2tqaJgSENaufsekAQUc55r5CAi0D14K8B5rIbgCr2wEgJoDGEQggoYZ7zaqvr/8JgOwwAAt979j36GRnNSOB+p8cXf1xQNfDvH5CbaS2GNpld4GWPRwQRKbTaispseuPAcPH/aJy8e4QekugPgIIah6+RC4PINu3Z1F8G3xNmTK1Wx+Hp2ck/O9FM1N39zAyc+ZMgunXO5uBdPQ5Uz0dy3UNv2HgIikpLCwj27b+RNcJDfUlSx+cg5L/b+tsO0wdO5ZBXnt9KfH19YPrN3dqG9t0gzbj4ibA86VThYUUWbXqbmNoaADhXK8ne6KbfTeQPi576ETe3t4kLHo0cfIIJTKtjiNVEhLgJiL/FxmX99mO5FmBfh3TRo2ZnOHr423zATJ8+HD6k01qiNoMaBy4tewHABwfmGtK3E2lsC7+DgFFoIGhwOhE7p+otb8m9IpANrOOCCYGgQQSSCCBekNioQsEEkgggQQSAEQggQQSSCABQAQSSCCBBBIARCCBBBJIIAFABBJIIIEEEshE/y/AACVIt4/lGbBHAAAAAElFTkSuQmCC");background-size:auto 60px}}.rendered_html ul{margin:0}</style><script src="https://ajax.googleapis.com/ajax/libs/jquery/3.1.0/jquery.min.js"></script><script>generateHtml=function(container){var containerID=container.attr('id');var preload=container.attr("preload")=="none"?false:true;var srcAttr='url';if(preload)srcAttr="src";var nbContainer=container.parent().attr("containers");if(nbContainer){nbContainer=parseInt(nbContainer)+1;container.parent().attr("containers",nbContainer);}else{container.parent().attr("containers",1);}
```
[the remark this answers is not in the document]
boazcogan: I do not recall modifying this one at all, I'll revert it.

tests/datasets/test_hooks.py
```diff
@@ -52,6 +53,27 @@ def test_dataset_hook_mix_source_folder(mix_source_folder):
...
+    assert np.allclose(sum(_sources).audio_data, data['mix'].audio_data)
+
+    # check that the duration are all the same length
+    _durations = [len(elem) for elem in _sources]
+    assert np.all(np.array(_durations + [len(data['mix'])]) == _durations[0]), f"source durations do not match." \
+                                                                               f"Source durations: {_durations + [len(data['mix'])]}"
+    # and that all of the durations are 3 seconds long
+    assert _durations[0] // 44_100 == 1.0, f"source duration does not match the target length:" \
+                                           f" duration={_durations[0] / 44_100}, Target={1.0}"
+
```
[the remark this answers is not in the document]
boazcogan: Yes, I agree. I don't think the code I had there actually does that but I'll come up with something that does.

nussl/datasets/hooks.py
boazcogan: I have not tested it with multiple threads yet, I'll be sure to do that. It would have been a race condition given the initial implementation where offset was a class attribute. Changing it to a function variable should limit any possible collisions to be on read which is perfectly fine.

nussl/datasets/hooks.py
boazcogan: passing kwargs allows these functions to interact with the AudioSignal object by passing along the desired parameters. In this specific case it lets the SalientExcerptMixSourceFolder._get_mix to read an audio file with a specified offset and duration.

nussl/datasets/hooks.py
ethman: Oh, yeah. Nice. I see what you did here. Yeah, that's a nice work around.

nussl/datasets/base_dataset.py
```diff
@@ -278,7 +278,17 @@ def _load_audio_file(self, path_to_audio_file, **kwargs):
         Returns:
             AudioSignal: loaded AudioSignal object of path_to_audio_file
         """
-        audio_signal = AudioSignal(path_to_audio_file, **kwargs)
+        # Allow for the Salient dataset object to have a padded signal if the duration of the read file is too short
+        if "padding_mode" in kwargs:
```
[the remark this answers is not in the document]
boazcogan: I thought it would be more concise to add it here but I see the issue. I moved it out to the Salient class in the most recent commit.

nussl/datasets/hooks.py
```diff
@@ -417,6 +421,168 @@ def process_item(self, item):
...
+            # TODO: logging.info()
+            print(f'Preprocessing data at {self.folder}.')
+
+        # TODO: Balanced dataset!
+        items = self.mix_src.get_items(self.folder)
+        metadata = []
+        for item in tqdm.tqdm(items, disable=not self.verbose):
+            mix, sources = self.mix_src.get_mix_and_sources(item)
+            target_src = sources[self.salient_src]
+            # Todo: Repair omission code
```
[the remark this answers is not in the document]
boazcogan: This todo is a reminder to myself to add a mode for skipping samples that are too short, it looks like I missed it. In any case its in the wrong spot so I'll remove it for now and add it back in later.

tests/datasets/test_hooks.py
```diff
@@ -7,6 +7,7 @@
 from nussl.datasets import transforms
 import tempfile
 import shutil
+import librosa
```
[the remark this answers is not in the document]
boazcogan: I needed it in the test I wrote to make sure the samples were salient, but I'm realized that test doesn't quite work. I expect I'll need it when I revise the test. For now it can be removed.

nussl/datasets/hooks.py
boazcogan: I've adjusted the wording.

nussl/datasets/hooks.py
```diff
@@ -417,6 +423,125 @@ def process_item(self, item):
...
+            Path is defined relative to folder. If None, all folders other than
+            mix_folder are treated as the source folders. Defaults to None.
+        ext (list, optional): Audio extensions to look for in mix_folder.
+            Defaults to ['.wav', '.flac', '.mp3'].
+        threshold_db (float, optional): the minimum relative loudness of the salient source measured in decibels
+            relative to the rms of the salient source. Defaults to -60.0.
+        segment_dur (float, optional): the duration of the desired audio clips in seconds. Defaults to 4.0.
+        hop_ratio (float, optional): size of the hops to use when computing the RMS. Defaults to 0.5.
+        verbose (bool, optional): suppress progress bar and logging for the dataset preprocessing. Defaults to False.
+        padding_mode (int, optional): Flag representing how padding should be handled for clips shorter than the desired
```
ethman, line 450: It might be more readable to make this parameter use strings instead of ints.

Also, I'm not sure we need all of these modes. We just want padding or no padding. If `pad=True` then pad with zeros, else truncate the signal. I'm not sure that "same padding" and "leave as is" make much sense in this context.

nussl/datasets/hooks.py
```diff
@@ -417,6 +421,168 @@ def process_item(self, item):
...
+            len(self.song_metadata),
+            sample_rate=sample_rate,
+            **kwargs
+        )
+
+    def _populate_metadata(self):
+        """Makes a metadata dict of salient start times for each source."""
+
+        if self.verbose:
+            # TODO: logging.info()
```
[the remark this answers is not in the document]
ethman: Good job. Now you can get rid of this TODO comment :-)

nussl/datasets/hooks.py
ethman: Actually: see comment below.

nussl/datasets/hooks.py
```diff
@@ -417,6 +423,125 @@ def process_item(self, item):
...
+        OnTheFly will pass itself as well as the item to this function"""
+        item = self.song_metadata[item]
+        mixsrc_item = item['mixsrc_item']
+        # Note that the onset needs to be passed to the AudioSignal class as the kwarg offset.
+        onset = item['start']
+        salient_excerpt = self.mix_src.process_item(mixsrc_item, offset=onset/self.sample_rate,
+                                                    duration=self.segment_dur)
+        return self._pad_output(salient_excerpt)
+
+    def _pad_output(self, excerpt):
```
ethman, line 531: Okay, I'm thinking that it's best not to support padding at all. It kind of makes the logic a complicated mess, so my feeling is that we should stick with just truncating. There are some edge cases, like "what if all of the audio files are too short?" and I think in that case, we should steer the user to another object (MixSourceFolder), or to change the parameters of this object. If there's a good use case for padding in the future then we can think about adding it. 

For now, just make sure the code supports truncating the last excerpt and maybe open an issue to talk about padding the last window.

ethman: So this means that in `_get_mix()` we always assume the salient_excerpt is the correct length. The easiest way to accomplish this is to add a little bit of extra logic in the for loop at line 513 to just check if the window is long enough: `if start + self.segment_dur <= mix.duration: # keep`

ethman (commented):
Overall looks good again. Getting closer. Keep iterating. I think we should nix the padding, as discussed below.

One nit pick is to make sure your code is following PEP8, specifically limiting the maximum line length to 79 chars: https://pep8.org/#maximum-line-length

nussl/datasets/hooks.py
```diff
@@ -417,6 +423,125 @@ def process_item(self, item):
...
+                 source_folders: List[str] = None,
+                 ext: List[str] = None,
+                 make_mix: bool = False,
+                 threshold_db: float = -60.0,
+                 segment_dur: float = 4.0,
+                 hop_ratio: float = 0.5,
+                 verbose: bool = False,
+                 padding_mode: int = 0,
+                 **kwargs):
+        assert sample_rate is not None, f"Please provide a sample rate for the dataset."
```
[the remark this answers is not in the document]
boazcogan: I honestly did not realize I could do this, yes adjusting the code now.

nussl/datasets/hooks.py
```diff
@@ -417,6 +427,146 @@ def process_item(self, item):
...
+    Given a dataset in the expected format, this class identifies and
+    partitions the folder into a dataset containing segment_dur second audio
+    clips that contain a desired salient source. See MixSourceFolder for more
+    details on expected directory structure.
+
+    Args:
+        folder (str): Location that should be processed to produce the
+            list of files.
+        salient_src (str): The name of the source that will be used to identify
+         salient samples in the dataset. e.g. ('drums')
```
ethman, line 442: Make sure all of the wrapped lines have the same indent. Some have 4 spaces (or 1 tab) others have 2. Follow the docstrings in the rest of this file. This will be important when we have to build the documentation website.

nussl/datasets/hooks.py
```diff
@@ -417,6 +427,146 @@ def process_item(self, item):
...
+            mix_folder are treated as the source folders. Defaults to None.
+        ext (list, optional): Audio extensions to look for in mix_folder.
+            Defaults to ['.wav', '.flac', '.mp3'].
+        threshold_db (float, optional): the minimum relative loudness of the
+         salient source measured in decibels relative to the rms of the salient
+         source. Defaults to -60.0.
+        segment_dur (float, optional): the duration of the desired audio clips
+         in seconds. Defaults to 4.0.
+        hop_ratio (float, optional): size of the hops to use when computing
+         the RMS. Defaults to 0.5.
```
ethman, line 457: Not just the RMS, but more importantly the Segment window.

nussl/datasets/hooks.py
```diff
@@ -417,6 +427,146 @@ def process_item(self, item):
...
+        verbose (bool, optional): suppress progress bar and logging for the
```
ethman, line 458: This language is a bit ambiguous. Does "Suppress progress bar..." mean that if this arg is True it will suppress the progress bar? For bools, I like to be explicit in the docs: `If True, this flag will blah blah. Else, blah blah.`

nussl/datasets/hooks.py
```diff
@@ -417,6 +427,146 @@ def process_item(self, item):
...
+        for item in tqdm.tqdm(items, disable=not self.verbose):
+            mix, sources = self.mix_src.get_mix_and_sources(item)
+            target_src = sources[self.salient_src]
+            target_audio = target_src.audio_data
+            salient_starts = utils.find_salient_starts(target_audio,
+                                                       self.segment_dur,
+                                                       self.hop_ratio,
+                                                       mix.sample_rate,
+                                                       self.threshold_db)
+            # this is a fairly cheap operation, no need to embed it within a
```
ethman, line 517: I don't think you need this comment

nussl/datasets/hooks.py
```diff
@@ -417,6 +427,146 @@ def process_item(self, item):
...
+            # conditional
+            balanced_samples.append((len(salient_starts), item,
+                                     salient_starts, mix.sample_rate))
+            for start in salient_starts:
+                if (start + self.segment_dur * mix.sample_rate) <= len(mix):
+                    metadata.append({
+                        'mixsrc_item': item,
+                        'start': start / mix.sample_rate
+                    })
+        # if the balance flag is set then balance the metadata
```
ethman, line 527: Same with this comment. In general, use inline comments to explain code excerpts that might be difficult to parse like complex logic, a seemingly unorthodox decision, or the expected contents of a data structure. Adding unnecessary comments can clutter up the code.

A more effective comment here would be like `# self._balance_set() will take care of the balancing logic`. But I'm still not convinced you need a comment here though...

nussl/datasets/hooks.py
```diff
@@ -417,6 +427,146 @@ def process_item(self, item):
...
+
+    def _get_mix(self, placeholder, item):
+        """OnTheFly closure. Gets called for every iteration to build a batch.
+        A placeholder param is necessary since OnTheFly will pass itself as
+        well as the item to this function"""
+        item = self.song_metadata[item]
+        mixsrc_item = item['mixsrc_item']
+        # Note that the onset needs to be passed to the AudioSignal
+        # class as the kwarg offset.
+        onset = item['start']
```
ethman, line 564: Why not rename this variable to `start = item['start']`? That way the code is clearer and we don't need this comment to explain what's happening.

ethman: As it is right now, this variable is referred to by three names in just 3 lines of code: `'start'` (in `item['start']`), `onset`, and `offset` (in the process_item kwarg). I'm not sure if there's enough context to name the variable `offset` here because there are a few layers of abstraction before the `offset` arg is used directly, so I think `start` conveys what's happening here well.

Sorry for the long winded explanation; I'm walking you through my thought process.

nussl/datasets/hooks.py
```diff
@@ -417,6 +427,146 @@ def process_item(self, item):
...
+        Args:
+            sample_counts: counts of the salient samples per audio files
+        Returns:
+        """
+        # TODO: Verify the method of balancing, option 1: oversample until
+        #  everything matches the max or option 2: oversample some under sample
+        #  others
+        avg_length = np.mean([elem[0] for elem in sample_counts])
+        metadata = []
+        for count, song, starts, sample_rate in sample_counts:
```
ethman, line 546: Nit: the AudioSignal might not be a song, so perhaps this is a confusing variable name. `item` would be better and more consistent with the rest of the object.

nussl/datasets/hooks.py
```diff
@@ -417,6 +427,146 @@ def process_item(self, item):
...
+            target_src = sources[self.salient_src]
+            target_audio = target_src.audio_data
+            salient_starts = utils.find_salient_starts(target_audio,
+                                                       self.segment_dur,
+                                                       self.hop_ratio,
+                                                       mix.sample_rate,
+                                                       self.threshold_db)
+            # this is a fairly cheap operation, no need to embed it within a
+            # conditional
+            balanced_samples.append((len(salient_starts), item,
```
ethman, line 519: Let's chat about this.

nussl/datasets/hooks.py
```diff
@@ -417,6 +427,146 @@ def process_item(self, item):
...
+    partitions the folder into a dataset containing segment_dur second audio
+    clips that contain a desired salient source. See MixSourceFolder for more
+    details on expected directory structure.
+
+    Args:
+        folder (str): Location that should be processed to produce the
+            list of files.
+        salient_src (str): The name of the source that will be used to identify
+         salient samples in the dataset. e.g. ('drums')
+        sample_rate (int, optional): the sampling rate for the audio files.
```
ethman, line 443: Nit: Make sure to use proper grammar in docstrings, including capitalizing the first word of each sentence.

ethman (commented):
Looking really good @boazcogan! Most of my comments are nit picks. I do have some opinions about how you should structure your code & logic for the balancing, but I'll pop into your stand up tomorrow to discuss.

PR #231, opened 2022-01-21
silence ffmpeg

faroit:
nussl jupyter audio embedding utils are great. When using mp3s (going through ffmpeg) it becomes pretty difficult to supress the warnings called by ffmpy. This PR silences ffmpeg and makes the notebooks cleaner

### DIFF
--- a/nussl/core/play_utils.py
+++ b/nussl/core/play_utils.py
@@ -13,6 +13,7 @@ import importlib_resources as pkg_resources
 from . import templates
 from .utils import _close_temp_files
 
+import os
 multitrack_template = pkg_resources.read_text(templates, 'multitrack.html')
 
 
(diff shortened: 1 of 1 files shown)


faroit:
Actually while at it. Maybe we can make the Bitrate an argument? 128 is a bit low for music

ethman:
Awesome thanks! I just kicked off the CI for you. Yes, please feel free to add the bitrate as an kwarg. Depending on how you add the bitrate kwarg you might have to update [this test](https://github.com/nussl/nussl/blob/471e7965c5788bff9fe2e1f7884537cae2d18e6f/tests/core/test_play_utils.py#L10). Let me know when you're ready and I can review & merge.

faroit:
@ethman i decided to switch to stempeg which is based on the excellent `ffmpeg-python` wrapper. That one behaves nicer with respect to error logging etc. stempeg isn't really necessary but its already a requirement for `musdb` so why not just reduce the amount of code here?

With this PR and relying on `ffmpeg-python` the codec is automatically inferred from the file extension, which - i think - is what most users expect. 

We could optionally add a codec parameter, but this might just get way to complicated to use. For the same reason. I decided to not offer the a bitrate argument because for other codec than mp3, the defaults sound pretty good. I bumped up the mp3 bitrate to 160kbps. 

Let me know if there is anything else left to do.

faroit:
btw. it would be nice change to player to use https://github.com/audiolabs/trackswitch.js or https://naomiaro.github.io/waveform-playlist/ but i have no experience with building jupyter widgets, and it looks like this has to minimized all in one html, file, right?

faroit:
@ethman after some tests, i bumped it up to 256kbits. Can allow to run CI?

faroit:
@ethman i tried the player with other formats, and I found that '.wav' doesn't really work well (produces clicks), '.flac' works nice, however.

